PR #10, opened 2016-01-13
Fixes "Finish" button that takes user back to first question of quiz

MattCheetham:
On iPad users were been taken back to the first quiz question when pressing finish.

Somewhere this finish button has been allowed back in on iPad where it was previously removed. I have now conditionally removed it based on whether the quiz completion page is being shown on iPad and is not currently presented.


simonmitchell:
Recommend changing the equality operator to a ">" otherwise if the user is dug down in content when they enter the quiz they won't get the dismiss quiz button

PR #16, opened 2016-04-12
Removes line of code which removes an asset file if it can't find the…

### LINKED ISSUE #12: Make sure App Thinning doesn't strip out 0.75x assets if no 1x or 2x exists
Some objects in the CMS can be broken meaning that only a 0.75x asset will be provided in the initial bundle. Due to the new App Thinning script these won't be included in the xcassets folder, resulting in a "Corrupt Delta" on the app looking for the next delta update. Any res assets we don't use shouldn't be discarded unless at least a 1x or 2x asset is copied to Bundle.xcassets to avoid this delta-update issue.

simonmitchell:
… resolution of it, and adds a block of code which either clears up 0.75x/1.5x assets or uses them in the XCAssets catalogue if no 1x/2x/3x was provided for said asset. Closes #12


MattCheetham:
Code has not been tested and no unit tests cover this. This is just a note to run tests or implement unit tests before we merge.

simonmitchell:
Has been tested now @MattCheetham :) after the changes made above

PR #24, opened 2016-05-31
Fix/linking

simonmitchell:
Links not beginning in https:// or http:// were crashing SFSafariViewController, decided to fix that by whacking https:// on the front of any links without either!

### DIFF
--- a/ThunderCloud/UINavigationController+TSCNavigationController.m
+++ b/ThunderCloud/UINavigationController+TSCNavigationController.m
@@ -225,9 +225,20 @@ static TSCLink *retryYouTubeLink = nil;
         NSOperatingSystemVersion iOS9 = (NSOperatingSystemVersion){9, 0, 0};
         if ([[NSProcessInfo processInfo] isOperatingSystemAtLeastVersion:iOS9]) {
             
-            SFSafariViewController *safariViewController = [[SFSafariViewController alloc] initWithURL:link.url];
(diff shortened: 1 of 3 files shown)


ThunderCloud/UINavigationController+TSCNavigationController.m
```diff
@@ -225,9 +225,20 @@ - (void)TSC_handleWeb:(TSCLink *)link
         NSOperatingSystemVersion iOS9 = (NSOperatingSystemVersion){9, 0, 0};
         if ([[NSProcessInfo processInfo] isOperatingSystemAtLeastVersion:iOS9]) {
             
-            SFSafariViewController *safariViewController = [[SFSafariViewController alloc] initWithURL:link.url];
-            safariViewController.delegate = self;
-            [self presentViewController:safariViewController animated:true completion:nil];
+            NSURL *url;;
+            
+            if ([[link.url.absoluteString substringToIndex:7] isEqualToString:@"http://"] || [[link.url.absoluteString substringToIndex:8] isEqualToString:@"https://"]) {
```
MattCheetham: Isn't this going to throw an exception if the URL is nil? we get a lot of empty links from the CMS.

simonmitchell: Good shout, will make sure it doesn't!

PR #25, opened 2016-05-31
Feature/collections

simonmitchell:
Adds superclass for collection cells so they can share some logic.


simonmitchell:
I tested this on the Storm Demo app, and all collection views worked flawlessly still on that :)

PR #27, opened 2016-06-30
Native Vectors

MattCheetham:
Storm has support for native pages and until now we have never attempted to have the vector of the app to be a native page. This pull request fixes the app not loading correctly by checking for a native page identifier then loading that page instead.

I've identified this as a bug with Steve but for now, we can roll this out to catch it until all CMS are updated to accommodate.


simonmitchell:
Looks fine to me!

PR #30, opened 2016-07-05
Adds named page property to list page

MattCheetham:
Some pages can be named in Storm to make them easier to find and also so that they can still be located if the client deletes them and creates a new page with the same name.

This is useful when doing list page overrides.

Let me know if you're happy with it then I'll merge :)


simonmitchell:
Yup, perfectly happy with this!

PR #41, opened 2017-05-16
Fix/compiler warnings

simonmitchell:


ThunderCloud/TSCSplitViewController.m
```diff
@@ -64,6 +64,14 @@ - (instancetype)init
     return self;
 }
 
+- (void)setMenuButton:(UIBarButtonItem *)menuButton
+{
+    _menuButton = menuButton;
+    if ([((UINavigationController *)self.detailViewController) respondsToSelector:@selector(viewControllers)]) {
+        ((UIViewController *)((UINavigationController *)self.detailViewController).viewControllers[0]).navigationItem.leftBarButtonItem = menuButton;
```
simonmitchell: Yeah it would, will tweak this now :)

ThunderCloud/TSCAuthenticationController.m
```diff
@@ -18,7 +18,8 @@ - (instancetype)init
 {
     if (self = [super init]) {
         
-        [TSCDeveloperModeController shared];
+        TSCDeveloperModeController *devModeController = [TSCDeveloperModeController shared];
+        devModeController.baseURL = devModeController.baseURL;
```
MattCheetham: Am I missing something here or do you literally set a URL to itself?

ThunderCloud/UINavigationController+TSCNavigationController.m
```diff
@@ -494,8 +496,10 @@ - (void)TSC_handleYouTubeVideo:(TSCLink *)link
                 if (quality) {
                     
                     //Present the video
-                    TSCMediaPlayerViewController *viewController = [[TSCMediaPlayerViewController alloc] initWithContentURL:[NSURL URLWithString:[[NSString stringWithFormat:@"%@&signature=%@", videoDictionary[quality][@"url"], videoDictionary[quality][@"sig"]] stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding]]];
                     
+                    TSCMediaPlayerViewController *viewController = [[TSCMediaPlayerViewController alloc] init];
+                    AVPlayer *video = [AVPlayer playerWithURL:[NSURL URLWithString:[[NSString stringWithFormat:@"%@&signature=%@", videoDictionary[quality][@"url"], videoDictionary[quality][@"sig"]] stringByRemovingPercentEncoding]]];
```
MattCheetham: Risk of crashing of [quality][@"sig"] doesn't exist. Same for [quality][@"url"]

ThunderCloud/TSCAuthenticationController.m
simonmitchell: Hmmm I'm not entirely sure why this is happening, you're right...

simonmitchell: Ahhh I remember, it's because `TSCAuthenticationController` has to be the controller which initialises `TSCDeveloperModeController` due to (I think) some race conditions. However just calling the method was causing a compiler warning, so I added in this line

joeltrew:
@simonmitchell lol Simon I was just about to push a fix to the dictionary safety issue

ThunderCloud/TSCAuthenticationController.m
simonmitchell: Hmmm I can't go back to a time when that shared call wasn't in the code, I don't see why it would be causing issues though?

simonmitchell:
@joeltrew haha my bad!

PR #43, opened 2017-06-13
Adds support for regional languages

MattCheetham:
Upgrades to the storm system for ATLAS mean that storm now needs to be able to support regional languages (Canadian French, British English, etc) as opposed to just "French" before like we had.

This works by loading the "base" pack (English), then loading the overrides/differences from the minor/regional language pack (British) and applying them over the top.

I have taken the liberty of writing the language controller from scratch and while it works for my use case currently it would be good to get it combed over by others as at this point I am sick of looking at it and I think there's probably room for improvements that I just can't work out. 

Focus your efforts on `StormLanguageController.swift`


ThunderCloud/StormLanguageController.swift
```diff
@@ -0,0 +1,410 @@
...
+            
+            for preferredLocale in _preferredLocales {
+                
+                if preferredLocale.languageCode == pack.locale.languageCode && preferredLocale.regionCode == pack.locale.regionCode {
+                    regionalLanguagePack = pack
+                    
+                    //Set the major language if it matches
+                    if let _languageCode = pack.fileName.components(separatedBy: "_").last {
+                        let languageOnlyLocale = Locale(identifier: _languageCode)
+                        majorLanguagePack = LanguagePack(locale: languageOnlyLocale, fileName: _languageCode)
```
simonmitchell: @MattCheetham here and a couple of lines below we're creating a LanguagePack with a filename which is just say `en.json` (The very last component of the actual filePath to the pack) is it important that this file exists and is a correct filePath? Or do you later re-create a full filePath and check it exists? :)

ThunderCloud/StormLanguageController.swift
```diff
@@ -0,0 +1,410 @@
...
+        }
+        
+        //Fall back to default if we need it
+        if finalLanguage.count == 0 {
+            
+            if let appFileURL = ContentController.shared.fileUrl(forResource: "app", withExtension: "json", inDirectory: nil) {
+            
+                let appJSON = try? JSONSerialization.jsonObject(withFile:appFileURL.path, options: [])
+            
+                if let _appJSON = appJSON as? [String: AnyObject], let packString = _appJSON["pack"] as? String, let packURL = URL(string: packString) {
```
simonmitchell: Should appJSON be `[String: Any]` rather than `[String: AnyObject]`? I'm never sure of the difference

ThunderCloud/StormLanguageController.swift
```diff
@@ -0,0 +1,410 @@
...
+                    
+                    guard let fileName = packURL.lastPathComponent.components(separatedBy: ".").first, let fullFilePath = ContentController.shared.fileUrl(forResource: fileName, withExtension: "json", inDirectory: "languages") else {
```
simonmitchell: Where we have lines like this when we're separating strings by characters and getting last path components e.t.c. It'd be nice to have an example file path (Or multiple) so we can picture which part of the string you're looking to get :)

ThunderCloud/StormLanguageController.swift
```diff
@@ -0,0 +1,410 @@
...
+//
+//  Created by Matthew Cheetham on 02/06/2017.
+//  Copyright © 2017 threesidedcube. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+@objc(TSCStormLanguageController)
+/// A subclass of ThunderBasic's `TSCLanguageController` which adds extra properties and methods to the controller more useful to Storm's needs
```
simonmitchell: Would be good to get some top-level documentation on how storm strings are formatted differently to the strings `Locale` uses (hence why you have to flip the components of the string), and the difference between a Language and a Region e.t.c.

ThunderCloud/StormLanguageController.swift
```diff
@@ -0,0 +1,410 @@
...
+                    guard let fileName = packURL.lastPathComponent.components(separatedBy: ".").first, let fullFilePath = ContentController.shared.fileUrl(forResource: fileName, withExtension: "json", inDirectory: "languages") else {
+                        return
+                    }
+                    
+                    self.languageDictionary = languageDictionary(for: fullFilePath.path)
+                    return
+                }
+            }
+        }
+
```
simonmitchell: Previously we had a load of fallbacks here, like falling back to the English pack, and then falling back to just any language, are we sure we're not going to need these any more?

ThunderCloud/TSCAccordionTabBarViewController.m
```diff
@@ -52,7 +52,7 @@ - (instancetype)initWithDictionary:(NSDictionary *)dictionary parentObject:(id)p
                 [typeDictionary setValue:@"NavigationTabBarViewController" forKey:@"type"];
                 
                 TSCNavigationTabBarViewController *navTabController = [[TSCNavigationTabBarViewController alloc] initWithDictionary:typeDictionary];
-                navTabController.title = TSCLanguageDictionary(tabBarItemDictionary[@"tabBarItem"][@"title"]);
+                navTabController.title = [[TSCStormLanguageController sharedController] stringForDictionary:(tabBarItemDictionary[@"tabBarItem"][@"title"])];
```
simonmitchell: Could we add some type safety in here whilst we're changing up the language controller? :)

ThunderCloud/StormLanguageController.swift
simonmitchell: This one may not need changes :)

ThunderCloud/StormLanguageController.swift
simonmitchell: Also may not need changes :)

ThunderCloud/StormLanguageController.swift
MattCheetham: The packs that we're looping through are generated from the files in their directory so storing them as file name here is safe because it *does* exist. We later use the ContentController methods to generate a full file path and safely load the file.

I *think* that answers your question

ThunderCloud/StormLanguageController.swift
MattCheetham: I put it as `[String: AnyObject]` as that's how `NSDictionary` works and this is backwards compatible with Objective-C. But looking at more now I think it would be fine to have `[String: Any]` as the raw dictionary is not exposed to Obj-C, just the lookup methods that return `String`

ThunderCloud/StormLanguageController.swift
MattCheetham: CHANGE APPROVED

ThunderCloud/StormLanguageController.swift
MattCheetham: Because the loading works completely differently now I'm unsure on what we should do.

The whole point of the "pack" parameter in the app.json is that it supplies the default language file for the fallback. If the app doesn't have that then we're in serious trouble anyway and have had some major publishing issue.

Is it worth opening up the languages directory again and just selecting the first file I see?

ThunderCloud/StormLanguageController.swift
simonmitchell: It certainly does :D

ThunderCloud/StormLanguageController.swift
simonmitchell: I'm not sure tbh :/ It's just a choice at this stage I guess, if the publish is already borked does it matter?

simonmitchell:
I approve of this PR :P

ThunderCloud/StormLanguageController.swift
```diff
@@ -0,0 +1,446 @@
...
+        if let overrideObject = UserDefaults.standard.object(forKey: "TSCLanguageOverride") as? Data, let _overrideLanguage = NSKeyedUnarchiver.unarchiveObject(with: overrideObject) as? TSCLanguage, let localeIdentifier = _overrideLanguage.languageIdentifier {
+            _preferredLocales.insert(Locale(identifier: localeIdentifier), at: 0)
+            overrideLanguage = _overrideLanguage
+        }
+        
+        return _preferredLocales
+    }
+    
+    /// The locales that are available in the language packs
+    private var availableLocales: [LanguagePack]? {
```
simonmitchell: Would it not make more sense to name this `availableLanguagePacks` as order not to confuse it as an array of `Locale` objects?

joeltrew: Good point I agree, I'll make the change now

PR #44, opened 2017-06-14
Fix/corrupt delta removal

simonmitchell:
As far as Me and Matt can see this is to fix an issue where corrupt deltas weren't removed which was causing further deltas to not download correctly. We're not 100% sure of this, but we need to make sure we're deleting the zipped delta in the same place we're saving it which this does! (See line 380)


MattCheetham:
@simonmitchell Did we establish this is now covered in other branches? Let's close if so.

simonmitchell:
@MattCheetham yes I believe it is 👍🏻

simonmitchell:
@MattCheetham I'm going to change this to be merged into Swift, and see if that makes any difference to the changed lines...

simonmitchell:
Okay, so we do still need some of it, because the current code is removing `bundle.tar.gz` files in `deltaDirectory` where they are never moved. But I'm fairly sure we don't want to call `removeBundle(in: deltaDirectory)` because bundles are only moved to there if they are valid?

MattCheetham:
Hmm. I'll review this in my free-ish time next week and check that delta removal works as expected.

simonmitchell:
Not sure what's going on with indentation here now, all looks okay in Xcode, but I've made a few changes to how things were working (stupidly) like passing the same parameter into the unpack function... Could also look at passing the directory into the `removeCorruptDeltaBundle` function rather than assuming the corrupt bundle is in `temporaryUpdateDirectory`

PR #46, opened 2017-07-31
Adds a call to relayout buttons after cell text lays out

joeltrew:
Fixes issue where text would overlap buttons on 5s and smaller devices


joeltrew:
@simonmitchell Everything works fine and loads at a good speed from my testing

PR #48, opened 2017-08-04
Fix/badge find

joeltrew:
Updates BadgeController find method + adds generics and nullability

### DIFF
--- a/ThunderCloud/TSCBadgeController.h
+++ b/ThunderCloud/TSCBadgeController.h
@@ -7,6 +7,7 @@
 //
 
 @class TSCBadge;
+@class TSCGridItem;
 
 #import <Foundation/Foundation.h>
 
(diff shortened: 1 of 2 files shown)


ThunderCloud/TSCBadgeController.h
```diff
@@ -21,41 +22,41 @@
...
  */
-- (BOOL)hasEarntBadgeWithId:(NSString *)badgeId;
+- (BOOL)hasEarntBadgeWithId:(nonnull NSString *)badgeId;
 
 /**
  Marks and saves a badge as earnt
  @param badgeId The unique id for the `TSCBadge`
  */
-- (void)markBadgeAsEarnt:(NSString *)badgeId;
+- (void)markBadgeAsEarnt:(nullable NSString *)badgeId;
```
MattCheetham: Shouldn't this be non-null? Is there a benefit of letting null be passed to mark a badge as earnt?

joeltrew: `- (void)markBadgeAsEarnt:(NSString *)badgeId
{
    if (badgeId && ![self hasEarntBadgeWithId:badgeId]) {`

We check if the badgeId is nil in the method body so i assumed there are some places where nil is passed in?

joeltrew: https://github.com/3sidedcube/iOS-ThunderCloud/blob/develop/ThunderCloud/TSCBadgeController.m#L105

MattCheetham: Yeah I guess it's a bad idea to nonnull it at this point and end up in trouble for passing in nil later on :/

PR #50, opened 2017-09-12
Feature/languagepack migration

joeltrew:
Some changes to the LanguageController, deprecated TSCLanguage in favour of LanguagePack to make things clearer, includes a migration between the two

### DIFF
--- a/ThunderCloud/StormLanguageController.swift
+++ b/ThunderCloud/StormLanguageController.swift
@@ -99,8 +99,6 @@ public class StormLanguageController: NSObject {
         return nil
     }
     
-    
-    
     //MARK: - Migration Methods for TSCLanguage
     func migrateToLanguagePackIfRequired() {
         
@@ -117,7 +115,6 @@ public class StormLanguageController: NSObject {
             }
         }
     }
-    
     
     /// Converts a TSCLanguage object into the new LanguagePack Format
     ///
@@ -171,7 +168,6 @@ public class StormLanguageController: NSObject {
             
             for pack in _availableLanguagePacks {
             
-
                 // Matches both language and region
                 if preferredLocale.languageCode == pack.locale.languageCode &&
                     pack.locale.regionCode != nil &&
@@ -187,8 +183,9 @@ public class StormLanguageController: NSObject {
                     
                     return (regionalLanguagePack: regionalLanguagePack, majorLanguagePack: majorLanguagePack)
                     
-                    // Only matches language
-                } else if preferredLocale.languageCode == pack.locale.languageCode, majorLanguagePack == nil {
+                    // Only matches language, and if majorLanguage has not already been set
+                } else if preferredLocale.languageCode == pack.locale.languageCode,
+                    majorLanguagePack == nil {
                     
                     //Set the major language if only the language matches. Major language pack always exists if a minor one exists
                     if let _languageCode = pack.locale.languageCode, let languageName = pack.fileName.components(separatedBy: "_").first {
@@ -277,7 +274,6 @@ public class StormLanguageController: NSObject {
                 }
             }
         }
-        
         
         self.languageDictionary = finalLanguage
     }
--- a/ThunderCloud/StormLanguageController.swift
+++ b/ThunderCloud/StormLanguageController.swift
@@ -27,8 +27,15 @@ public class StormLanguageController: NSObject {
     /// The current language identifier
     public var currentLanguage: String?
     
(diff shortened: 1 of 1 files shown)


ThunderCloud/StormLanguageController.swift
```diff
@@ -38,17 +45,33 @@ public class StormLanguageController: NSObject {
...
-        //Add override locales if they exist
-        if let overrideObject = UserDefaults.standard.object(forKey: "TSCLanguageOverride") as? Data, let _overrideLanguage = NSKeyedUnarchiver.unarchiveObject(with: overrideObject) as? TSCLanguage, let localeIdentifier = _overrideLanguage.languageIdentifier {
-            _preferredLocales.insert(Locale(identifier: localeIdentifier), at: 0)
-            overrideLanguage = _overrideLanguage
+        // If the user has applied an override language to the app we need to retrieve a saved version and apply it as the app language
+        // Check the defaults for the override language filename
+        if let overridePackFileName = UserDefaults.standard.object(forKey: overrideLanguagePackSavingKey) as? String {
+            
+            // If we have the saved override filename, filter our available language packs for it
+            let savedOverridePack = availableLanguagePacks?.filter({ (pack) -> Bool in
```
simonmitchell: You can just do a `first(where: {})` here :)

ThunderCloud/StormLanguageController.swift
```diff
@@ -38,17 +45,33 @@ public class StormLanguageController: NSObject {
...
     
+    // Private init as only the shred instance should be used
+    private override init() {
+        super.init()
+        migrateToLanguagePackIfRequired()
+    }
+    
     /// The locales that are available in the language packs
-    private var availableLocales: [LanguagePack]? {
+    public var availableLanguagePacks: [LanguagePack]? {
```
simonmitchell: 👍

ThunderCloud/StormLanguageController.swift
```diff
@@ -76,6 +99,43 @@ public class StormLanguageController: NSObject {
...
     
+    //MARK: - Migration Methods for TSCLanguage
+    func migrateToLanguagePackIfRequired() {
+        
+        //Add override locales if they exist
+        if let overrideObject = UserDefaults.standard.object(forKey: "TSCLanguageOverride") as? Data, let _overrideLanguage = NSKeyedUnarchiver.unarchiveObject(with: overrideObject) as? TSCLanguage {
+            
+            // Migrate the languageOverride to languagePack
+            if let pack = languagePack(for: _overrideLanguage) {
+                UserDefaults.standard.set(pack.fileName, forKey: "TSCLanguagePackOverrideFileName")
```
simonmitchell: Use `overrideLanguagePackSavingKey` here rather than create the string again

ThunderCloud/StormLanguageController.swift
```diff
@@ -95,19 +155,24 @@ public class StormLanguageController: NSObject {
     private func languagePacks() -> (regionalLanguagePack: LanguagePack?, majorLanguagePack: LanguagePack?)? {
         
         //Find out if any locales match
-        guard let _availableLocales = availableLocales, let _preferredLocales = preferredLocales else {
+        guard let _availableLanguagePacks = availableLanguagePacks, let _preferredLocales = preferredLocales else {
```
simonmitchell: I've been trying to move away from this `_` syntax when not needed, in this case there's no reason why they're needed :) Sometimes they are, and when they are we should use the underscore :D

joeltrew: I agree 100%, I almost always shadow the variable but I'll be honest I did a find and replace for this 😅

ThunderCloud/StormLanguageController.swift
joeltrew: Not sure why I didn't do this 🤦‍♂️

ThunderCloud/StormLanguageController.swift
joeltrew: 👌

ThunderCloud/StormLanguageController.swift
simonmitchell: 😂 probably best do do a cmd+f for the rest of the file too, in case I missed any :)

ThunderCloud/StormLanguageController.swift
simonmitchell: Hahaaa 😝 makes sense now! Yeah when I re-wrote Storm in swift I changed all of the `_` to just shadow the variable 👍

PR #51, opened 2017-09-14
Changes files:inDictionary on ContentController to only return unique fileNames

joeltrew:
The current implementation of the method checks the bundle and delta update folder for files, however if the file exists in both directories, the filename will appear twice in the array which I don't think is something you would ever want. Let me know :)

### DIFF
--- a/ThunderCloud/ContentController.swift
+++ b/ThunderCloud/ContentController.swift
@@ -1049,17 +1049,17 @@ public extension ContentController {
     ///
     /// - parameter inDirectory: The directory to look for files in
     ///
-    /// - returns: An array of file names
-    public func files(inDirectory: String) -> [String]? {
+    /// - returns: A set of file names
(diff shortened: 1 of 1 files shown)


joeltrew:
Does it make a difference which order? It checks both and if the file exists in either it will appear in the set? No?

simonmitchell:
Well, the one in the delta update should be preferred over the one in the initial bundle as this is the latest content :D The file paths are different whether they're in the bundle or delta bundle

joeltrew:
But this method doesn't return the filePaths, just the fileNames so it doesn't matter the order

simonmitchell:
Never mind, I'm an idiot... forgot `contentsOfDirectory(atPath:` doesn't return the full path!

joeltrew:
Renamed to fileNames to make it clearer

PR #52, opened 2017-09-19
Feature/localisable links

joeltrew:
Adds localised links to TSCLink

### DIFF
--- a/ThunderCloud/TSCLink+Localised.swift
+++ b/ThunderCloud/TSCLink+Localised.swift
@@ -16,7 +16,7 @@ public extension TSCLink {
     ///
     ///
     /// - Parameter dictionary: The initialisation dictionary passed from initWithDictionary
-    public func linkFromLocalisedLink(with dictionary: [AnyHashable: Any]) {
+    public func localised(with dictionary: [AnyHashable: Any]) {
         
         guard let urlDictionaries = dictionary["links"] as? [[AnyHashable: Any]] else { return }
         
--- a/ThunderCloud/TSCLink+Localised.swift
+++ b/ThunderCloud/TSCLink+Localised.swift
@@ -7,3 +7,58 @@
 //
 
 import Foundation
+
+public extension TSCLink {
+    
+    public func linkFromLocalisedLink(with dictionary: [AnyHashable: Any]) {
+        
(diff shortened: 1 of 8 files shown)


ThunderCloud/TSCLink+Localised.swift
```diff
@@ -0,0 +1,107 @@
...
+
+public extension TSCLink {
+    
+    /// Helper to initialise a localised link. This is called in TSCLink's initWithDictionary. The method will attempt to correctly initialise the link's url property from the dictionary by matching against the current language.
+    ///
+    /// The method will attempt to find a full match first i.e if the users language is usa_eng it will try to match that, failing that it will fall back to basic eng, and finally fall back to the first url in the links array.
+    ///
+    ///
+    /// - Parameter dictionary: The initialisation dictionary passed from initWithDictionary
+    public func linkFromLocalisedLink(with dictionary: [AnyHashable: Any]) {
```
simonmitchell: I would expect this to return something given the method name :) I'd rename this to something more suitable maybe, like:

`localise(with dictionary: [AnyHashable: Any])`

ThunderCloud/TSCLink+Localised.swift
```diff
@@ -0,0 +1,107 @@
...
+        guard let urlDictionaries = dictionary["links"] as? [[AnyHashable: Any]] else { return }
+        
+        guard let urlType = dictionary["type"] as? String else { return }
+        
+        self.linkClass = urlType
+        
+        let urls = urlDictionaries.flatMap({ LocalisedLinkContents(from: $0) })
+        
+        
+        func findLinkForLocale(using urls: [TSCLink.LocalisedLinkContents]) -> TSCLink.LocalisedLinkContents? {
```
simonmitchell: Any reason why this function is created in the scope of it's containing function? :P

joeltrew: It's only used inside that function, I could move out outside but I can't see anywhere else it would be called

ThunderCloud/TSCLink+Localised.swift
joeltrew: Whoops I meant to use localise not localised! Fixed, thanks Simon 👍

PR #53, opened 2017-11-17
Feature/iOS11 upgrades

simonmitchell:


ThunderCloud/TSCAppDelegate.swift
```diff
@@ -0,0 +1,196 @@
...
+
+	open func handleNotificationResponse(for notification: UNNotification, response: UNNotificationResponse?, fromLaunch: Bool) -> Bool {
+		
+		// Make sure the action wasn't dismissing the notification
+		guard response?.actionIdentifier != UNNotificationDismissActionIdentifier else {
+			return false
+		}
+		
+		guard let payload = notification.request.content.userInfo["payload"] as? [AnyHashable : Any] else { return false }
+		guard let url = payload["url"] as? String else { return false }
```
ryanbourneuk: Could combine this with the guard above, as they're both returning false :)

ThunderCloud/TSCAppDelegate.swift
```diff
@@ -0,0 +1,196 @@
...
+		
+		// Call handleNotificationResponse(for:response:fromLaunch) every second if we are presenting
+		if window?.rootViewController?.presentedViewController != nil {
```
ryanbourneuk: I suppose this could also be a guard too, so that it handles `return handleContentNotificationFor(cacheURL: url)` first. Keeps the golden path neat, as above :)

ThunderCloud/TSCAppDelegate.swift
```diff
@@ -0,0 +1,196 @@
...
+			
+			return handleContentNotificationFor(cacheURL: url)
+		}
+		
+		return true
+	}
+	
+	open func handleContentNotificationFor(cacheURL: String) -> Bool {
+		
+		if Bundle.main.infoDictionary?["TSCStreamingBaseURL"] as? String != nil {
```
ryanbourneuk: I'm unsure about what could be done here, whether perhaps each part of this if-statement should in fact be moved out to two separate functions, which get the `cacheURL` passed into them. That way, it reduces the level of indentation, especially as there's blocks in use, and helps with self-documenting the code (and documentation is pretty awesome!) Even further, there looks like there may be some code that can be reused across both flows of the statement.  What are your opinions? :)

ThunderCloud/TSCAppDelegate.swift
```diff
@@ -0,0 +1,196 @@
...
+	}
+	
+	open func userNotificationCenter(_ center: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
+		
+		if let foregroundOptions = foregroundNotificationOptions {
+			completionHandler(foregroundOptions)
+			return
+		}
+		
+	    _ = handleNotificationResponse(for: notification, response: nil, fromLaunch: false)
```
ryanbourneuk: Could this be shifted along to be in line with the rest of the function please? :)

ThunderCloud/TSCAppDelegate.swift
```diff
@@ -0,0 +1,196 @@
...
+	
+	open func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([Any]?) -> Void) -> Bool {
+		
+		guard let searchableItemIdentifier = userActivity.userInfo?[CSSearchableItemActivityIdentifier] as? String, searchableItemIdentifier.contains(".json") else {
+			return false
+		}
+		
+		guard let url = URL(string: "caches://pages/\(searchableItemIdentifier)") else { return false }
+		
+		guard let stormViewController = TSCStormViewController(url: url) else { return false }
```
ryanbourneuk: Could probably combine this guard with the one above, both return false and `url` from the first one is used here. Unsure whether it should all be rolled into the top-most guard, as then there'd be lots going on in just one guard and it'd be a bit unwieldy. I've often used a limit of 3 conditions per guard to keep it clean, what is your opinion on it? :)

ThunderCloud/TSCAppDelegate.swift
```diff
@@ -0,0 +1,196 @@
...
+//  Copyright © 2017 threesidedcube. All rights reserved.
+//
+
+import UIKit
+import UserNotifications
+import ThunderRequest
+
+@UIApplicationMain
+/// A root app delegate which sets up your window and push notifications e.t.c.
+open class TSCAppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterDelegate {
```
ryanbourneuk: For `UNUserNotificationCenterDelegate`, the functions here that conform to that delegate could perhaps be moved to an extension of `TSCAppDelegate` that conforms specifically to that delegate. Will help with self-documenting the code too! :)

ThunderCloud/TSCAppDelegate.swift
ryanbourneuk: Actually, if this is being recalled once per second while we're presenting the view controller, we could perhaps check `window?.rootViewController?.presentedViewController != nil` first before unwrapping the payload, as otherwise it isn't being used. What do you think? :)

ThunderCloud/TSCAppDelegate.swift
simonmitchell: Do you know whether this works with subclassing? I know in Objective-C it was hit-and-miss as to which methods would be called if you then subclassed `TSCAppDelegate` and wanted to override any of the delegate methods, but unsure about swift...

ThunderCloud/TSCAppDelegate.swift
simonmitchell: That's fine with me, I will try and move these changes over to the `feature/swift` branch as well 👍

simonmitchell: Actually scrap that, sorry @ryanbourneuk the reason these are above the check for presented view controller is because we don't even need to re-call once per second if there is no payload url. Will move them into one guard statement but they need to stay above the rest of the method

MattCheetham:
@simonmitchell Again, is this now covered in other branches?

ThunderCloud/TSCAppDelegate.swift
simonmitchell: I've tried to avoid combining basically when it's going to make the line significantly longer than other lines around it. however I think in this case it should be fine 😀

simonmitchell:
@MattCheetham it is yes, but still PR'ing for a couple of reasons: ARC Emergency is still using objective-c ThunderTable but has this change in it (for the benefit of using `UNUserNotificationCenter`) and it's also a way to PR part of the massive `feature/swift` branch without you guys having to trawl through a million changed lines of code

ThunderCloud/TSCAppDelegate.swift
simonmitchell: I'd like to keep it as-is for now. It's unlikely we'll be testing this because it's heavily UI based and reliant on network usage which isn't easy to do in test suites AFAIK. But mainly I'd like to keep it as-is because I want to keep the App Delegate as clean as possible, with as few functions as possible. Ideally the App Delegate should only have delegate methods for `UIApplicationDelegate` in it, we need to investigate something @MattCheetham shared with us a while back to try and clean up the `AppDelegate` a bit, so I'm okay with this being as is for a while 😊

ThunderCloud/TSCAppDelegate.swift
ryanbourneuk: From my thinking, if you create an extension of `TSCAppDelegate` (so not necessarily subclassing it), it should still call everything correctly. My recommendation would be to give it a shot to see if it does work :)

ThunderCloud/TSCAppDelegate.swift
ryanbourneuk: Ah yes! That's very true indeed! As you were :)

PR #54, opened 2017-11-17
Fix/navigation styling

simonmitchell:


simonmitchell:
@MattCheetham @ryanbourneuk suggesting that this be deleted as it's such minor changes that I believe are already made on the `feature/swift` branch

PR #55, opened 2017-12-12
Implements os_log for ContentController

MattCheetham:
Implemented the new os_log stuff for the ContentController.

Mostly just requires sanity checking that I've used the right log type (info, debug, error, fault) on error messages.

As a side note I would like to adopt this moving forward for all logging.

### DIFF
--- a/ThunderCloud/NavigationController.swift
+++ b/ThunderCloud/NavigationController.swift
@@ -21,6 +21,11 @@ public extension UINavigationController {
 	/// - Parameter link: A `StormLink` to decide which action to perform
 	@objc public func push(link: StormLink) {
 		
+		if let appDelegate = UIApplication.shared.delegate as? TSCAppDelegate, !appDelegate.linkIsWhitelisted(link) {
+			print("[Storm] Tried to push \(link.url?.absoluteString ?? "??") which is not a whitelisted link")
+			return
+		}
+		
 		let pathExtension = link.url?.pathExtension
 		let scheme = link.url?.scheme
 		let host = link.url?.host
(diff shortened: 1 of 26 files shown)


simonmitchell:
@MattCheetham can we update this to be pulling into `feature/swift`?

ThunderCloud/DeveloperModeController.swift
```diff
@@ -127,16 +127,17 @@ public class DeveloperModeController: NSObject {
...
-            
-            if let progressHandler = progressHandler {
-                progressHandler(stage, downloaded, totalSize, error)
-            }
-            
-            if stage == .finished {
-                DeveloperModeController.appIsInDevMode = true
-                self?.finishSwitching()
-            }
+        if let _deltaDirectory = ContentController.shared.deltaDirectory {
```
ryanbourneuk: Wondering if it'd be nicer to have this as a guard instead, just to reduce the levels of indentation going on. What do you think? :)

ThunderCloud/NavigationController.swift
```diff
@@ -70,7 +70,8 @@ public extension UINavigationController {
 			
 			if let splitViewController = UIApplication.shared.keyWindow?.rootViewController as? SplitViewController, UI_USER_INTERFACE_IDIOM() == .pad {
 				
-				splitViewController.show(viewController, sender: self)
+                splitViewController.setRightViewController(viewController, from: self)
```
ryanbourneuk: Indentation seems a bit off here, and is the line below still required?

PR #58, opened 2018-01-18
Checks for direct linkClass when handling page

simonmitchell:
Before swift re-write this method didn't use `else if` so the later statement checking `if link.linkClass == .app` would still be run and the app switch dialogue shown. However now that I've changed it to `if else` this check needs to be more stringent... The only linkClass that should be handled as pushing a cached json file is `internal`

### DIFF
--- a/ThunderCloud/NavigationController.swift
+++ b/ThunderCloud/NavigationController.swift
@@ -38,7 +38,7 @@ public extension UINavigationController {
 			
 			handleITunes(url: url)
 			
-		} else if pathExtension == "json" || scheme == "app" && link.linkClass != .native {
+		} else if (pathExtension == "json" || scheme == "app") && link.linkClass == .internal {
 			
 			handlePage(link: link)
 			


simonmitchell:
@MattCheetham This breaks creating links and trying to push them manually with url strings such as: 
```
guard let nextQuizLink = URL(string: "cache://pages/\(nextQuizID).json") else {
	return nil
}
return StormLink(pageId: nextQuizID)
```
because these links do not have the `linkClass` set to `app`... so if we merge this it will require that when constructing page links we use the methods: `StormLink(pageId:)` or `StormLink(pageName:)` from now on. Are we okay with that?

MattCheetham:
Hmm. I mean my gut says yes that is fine but I'm imagining there are going to be situations where this isn't suitable. Is there no way to set the linkClass based on the init of `StormLink(url:)` ?

simonmitchell:
That was another thought I had, I could check the url's scheme and root and if they are `cache` and `pages` respectably I can set the `linkClass` to the correct value? I'll work something up!

PR #61, opened 2018-02-06
Feature/swift quizzes

simonmitchell:
Updates quizzes to use a better navigation method, be written in swift, and be easier to override.

Please note a lot of the changed files are adding `ThunderBasics` and `ThunderTable` imports because when quiz imports were removed from the framework header file suddenly they needed importing in all the swift files.


ThunderCloud/QuizQuestion.swift
```diff
@@ -0,0 +1,85 @@
...
+		
+		if let winDictionary = dictionary["win"] as? [AnyHashable : Any] {
+			winText = StormLanguageController.shared.string(for: winDictionary)
+		} else {
+			winText = nil
+		}
+	}
+	
+	func reset() {
+
```
MattCheetham: Three empty methods?

simonmitchell: Aye @MattCheetham I guess really `QuizQuestion` should be a protocol, they're methods used by the subclasses of `QuizQuestion` 👍🏻 What do you think? Protocol based, or just document them up? How likely are we to have a new question type?

ThunderCloud/QuizQuestionContainerViewController.swift
```diff
@@ -0,0 +1,267 @@
...
+		progressView.transform = progressView.transform.concatenating(CGAffineTransform(scaleX: 1.0, y: 3.0))
+		
+		progressContainer.addSubview(progressView)
+			
+		let progressStartCap = UIView(frame: CGRect(x: progressView.frame.minX - 2, y: progressView.frame.minY, width: 6, height: 6))
+		progressStartCap.cornerRadius = 3.0
+		progressStartCap.backgroundColor = progressView.progressTintColor
+		progressContainer.addSubview(progressStartCap)
+		
+		let progressEndCap = UIView(frame: CGRect(x: progressView.frame.maxX - 3, y: progressView.frame.minY, width: 6, height: 6))
```
ryanbourneuk: Is there any reason why it's `-3` here, and `-2` above?

ThunderCloud/QuizImageSelectionViewController.swift
```diff
@@ -0,0 +1,103 @@
+//
+//  QuizImageSelectionViewController.swift
+//  GNAH
```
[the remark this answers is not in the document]
simonmitchell: Yes! Silly Copy Pasta

ThunderCloud/QuizQuestionContainerViewController.swift
simonmitchell: Not entirely sure, it was taken from the old objective-c code...

simonmitchell:
@ryanbourneuk Fixed most of those issues 😄

ThunderCloud/QuizQuestion.swift
MattCheetham: I haven't had time to go through fully but what you describe sounds like a prime opportunity for a protocol.

simonmitchell:
@MattCheetham Merged `feature/swift` in so should be no more merge conflicts 👍

PR #62, opened 2018-02-07
Changing how overridden languages are used.

ryanbourneuk:
Previously, the flow for using overridden languages was baked into the same code that handled detecting the device's language. This was affecting how overridden languages were being used - in some cases it would ignore the overridden language in preference of another. A previous fix resolved that issue, but also impacted upon how the device's regional version of a language was picked up. This merge request aims to solve both issues - resolving how overridden languages are used by moving the flow out of the standard detection flow, and fixing how the language detection works too.

I've also added further documentation to the codebase to clarify the purpose of some parts, and shifted some reusable code out into smaller functions.

Any questions, let me know!

### DIFF
--- a/ThunderCloud/StormLanguageController.swift
+++ b/ThunderCloud/StormLanguageController.swift
@@ -240,17 +240,20 @@ public class StormLanguageController: NSObject {
         // If we have an override set to use, lets try to use that instead of detecting language packs here.
         // If we can't use it, then we'll fallback to use normal detection.
         if let overrideLanguagePack = overrideLanguagePack {
-            let overrideFileName = overrideLanguagePack.fileName
+            let fileName = overrideLanguagePack.fileName
             
-            currentLanguage = overrideFileName
+            currentLanguage = fileName
             
             // Handle the major language element of this language (i.e., eng from usa_eng). If there is no region specified, this will just return the language.
-            if let majorLanguage = overrideFileName.components(separatedBy: "_").last, let majorPackPath = ContentController.shared.fileUrl(forResource: majorLanguage, withExtension: "json", inDirectory: "languages") {
+            if let majorLanguage = fileName.components(separatedBy: "_").last,
+                let majorPackPath = ContentController.shared.fileUrl(forResource: majorLanguage, withExtension: "json", inDirectory: "languages") {
                 writeLanguageData(from: majorPackPath.path, into: &finalLanguage)
                 
-                // Now attempt to handle the regional version of the language (i.e., usa_eng). If there was no region specified and we've already handled this above, we skip this as it's unnecessary.
-                if let regionalLanguagePath = ContentController.shared.fileUrl(forResource: overrideFileName, withExtension: "json", inDirectory: "languages"), majorLanguage != overrideFileName {
-                    writeLanguageData(from: regionalLanguagePath.path, into: &finalLanguage)
+                // Now attempt to handle the regional version of the language (i.e., usa_eng).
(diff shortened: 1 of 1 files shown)


ThunderCloud/StormLanguageController.swift
```diff
@@ -217,45 +216,71 @@ public class StormLanguageController: NSObject {
...
+            let overrideFileName = overrideLanguagePack.fileName
+            
+            currentLanguage = overrideFileName
+            
+            // Handle the major language element of this language (i.e., eng from usa_eng). If there is no region specified, this will just return the language.
+            if let majorLanguage = overrideFileName.components(separatedBy: "_").last, let majorPackPath = ContentController.shared.fileUrl(forResource: majorLanguage, withExtension: "json", inDirectory: "languages") {
+                writeLanguageData(from: majorPackPath.path, into: &finalLanguage)
+                
+                // Now attempt to handle the regional version of the language (i.e., usa_eng). If there was no region specified and we've already handled this above, we skip this as it's unnecessary.
+                if let regionalLanguagePath = ContentController.shared.fileUrl(forResource: overrideFileName, withExtension: "json", inDirectory: "languages"), majorLanguage != overrideFileName {
```
simonmitchell: The `majorLanguage != overrideFileName` detects here if I'm correct that the currentLanguage is just a base language, for example `en` rather than `en_us` correct? Maybe put that in the second sentence of the comment as took me a little while to work out why that was there! 😊

ThunderCloud/StormLanguageController.swift
```diff
@@ -208,7 +208,6 @@ public class StormLanguageController: NSObject {
                     //Set the major language if only the language matches. Major language pack always exists if a minor one exists
                     if let languageCode = pack.locale.languageCode, let languageName = pack.fileName.components(separatedBy: "_").first {
                         majorLanguagePack = LanguagePack(locale: Locale(identifier: languageCode), fileName: languageName)
-                        return (regionalLanguagePack: nil, majorLanguagePack: majorLanguagePack)
```
simonmitchell: Does removing this not mean we could get landed with a major language pack of, say `en` but a minot language pack of say: `es` so `es_en` with overrides taken from the Spanish regional pack?

MattCheetham:
Currently doing a write-up on expected language logic to make sure we are all on the same page. Going to hold off on reviewing this until I've finished sorting. Hopefully in the next hour!

ThunderCloud/StormLanguageController.swift
MattCheetham: Yeah I think you might be right @simonmitchell

MattCheetham:
@ryanbourneuk I don't have time to properly test until next week in my free time. If this fixes the issue for you on your project then build from this for now and then I can update you next week once I've sorted everything out.

ThunderCloud/StormLanguageController.swift
ryanbourneuk: So this line here is actually one I added in on Monday which was causing an issue with the region detection. If you check back to be1460677266c64f5c2365c3174f8cfff1f90335 you'll see where I've added it :)

ThunderCloud/StormLanguageController.swift
ryanbourneuk: Will make a note of that in the comment. That is indeed correct :)

ryanbourneuk: @simonmitchell It seems the documentation was there, but as it was inline with the first sentence it seemed quite unclear. Shifted it around, and renamed a few vars in 32d708b20079acc3487949b863ba462245b06852 :)

PR #63, opened 2018-02-12
Localisation Parser

MattCheetham:
Pull request created on behalf of @simonmitchell.

Hopefully he can give more info so we can get this closed off.


simonmitchell:
This is a parser for extracting all localised strings from a project. For now it needs to be run as a project rather than run from terminal, but Ideally we'd merge it in and then that can be done at a later date.

PR #76, opened 2018-02-16
Feature/gridpage swift rewrite

simonmitchell:
This has been re-written so we can use the new ThunderCollection library moving forward, and to improve overridability in projects.


ThunderCloud/GridItem.swift
```diff
@@ -0,0 +1,69 @@
...
+	open let description: String?
+	
+	public required init?(dictionary: [AnyHashable : Any]) {
+		
+		remainSelected = false
+		
+		if let linkDicationary = dictionary["link"] as? [AnyHashable : Any] {
+			link = StormLink(dictionary: linkDicationary)
+		} else {
+			link = nil
```
ryanbourneuk: Won't this already be nil, do we need to have an `else` clause for each item? :)

ThunderCloud/GridPage.swift
```diff
@@ -0,0 +1,105 @@
...
+//  Copyright © 2018 threesidedcube. All rights reserved.
+//
+
+import ThunderCollection
+import ThunderTable
+
+/// A subclass of `TSCCollectionViewController` for displaying a CMS grid page
+open class GridPage: CollectionViewController, StormObjectProtocol {
+	
+	/// An array of dictionaries which contain custom attributes fot the `StormObject`
```
ryanbourneuk: `for` instead of `fot` :)

MattCheetham:
Please remember to fill out a little description for the pull request. Hard to know exactly why things were re-written or where they were used when we weren't involved :)

ThunderCloud/GridItem.swift
```diff
@@ -0,0 +1,69 @@
...
+			description = StormLanguageController.shared.string(for: subtitleDict)
+		} else {
+			description = nil
+		}
+		
+		image = StormGenerator.image(fromJSON: dictionary["image"])
+	}
+	
+	public func configure(cell: UICollectionViewCell, at indexPath: IndexPath, in collectionViewController: CollectionViewController) {
+
```
MattCheetham: Empty method? Remove if not needed

ThunderCloud/GridPage.swift
```diff
@@ -0,0 +1,105 @@
...
+	public var attributes: [[AnyHashable : Any]]?
+	
+	private let dictionary: [AnyHashable : Any]
```
MattCheetham: Could you document this? It's unusual that we store a dictionary for a storm object so would be good to write up why we do.

ThunderCloud/GridPage.swift
```diff
@@ -0,0 +1,105 @@
...
+		
+		if let titleDict = dictionary["title"] as? [AnyHashable : Any], let titleContentKey = titleDict["content"] as? String {
+			title = StormLanguageController.shared.string(forKey: titleContentKey)
+		}
+	}
+	
+	required public init?(coder aDecoder: NSCoder) {
+		
+		dictionary = [:]
+		pageId = nil
```
MattCheetham: Rather than do this would it be better to give the properties a default value in their declaration?

ThunderCloud/QuizGridCell.swift
```diff
@@ -0,0 +1,14 @@
...
+//
+//  Created by Simon Mitchell on 16/02/2018.
+//  Copyright © 2018 threesidedcube. All rights reserved.
+//
+
+import UIKit
+
+/// Legacy equivalent of `QuizGridItem`
+class QuizGridCell: QuizGridItem {
+
```
MattCheetham: Move up the brackets to be on one line if it's just a dummy implementation like this

ThunderCloud/QuizGridItem.swift
```diff
@@ -0,0 +1,33 @@
...
+	let badgeId: String?
+
+	public required init?(dictionary: [AnyHashable : Any]) {
+		
+		if let badgeId = dictionary["badgeId"] as? String {
+			self.badgeId = badgeId
+		} else if let badgeId = dictionary["badgeId"] as? Int {
+			self.badgeId = "\(badgeId)"
+		} else {
+			badgeId = nil
```
MattCheetham: `badgeId` is nil by default so this else is not required

ThunderCloud/StandardGridCell.swift
```diff
@@ -0,0 +1,14 @@
...
+//
+//  Created by Simon Mitchell on 16/02/2018.
+//  Copyright © 2018 threesidedcube. All rights reserved.
+//
+
+import UIKit
+
+/// Legacy equivalent of `StandardGridItem`
+class StandardGridCell: StandardGridItem {
+
```
MattCheetham: Move brackets to single line

ThunderCloud/StandardGridItemCell.swift
```diff
@@ -0,0 +1,19 @@
...
+//  ThunderCloud
+//
+//  Created by Simon Mitchell on 16/02/2018.
+//  Copyright © 2018 threesidedcube. All rights reserved.
+//
+
+import UIKit
+import ThunderTable
+
+open class StandardGridItemCell: UICollectionViewCell {
```
MattCheetham: Document this up and give a brief overview of how it's displayed visually.

ThunderCloud/GridItem.swift
simonmitchell: Unfortunately not, because it's an optional let, it needs to have it's value set in init() method otherwise it won't compile!

ThunderCloud/GridItem.swift
simonmitchell: This needs to be implemented otherwise for some reason the function doesn't get called on subclasses... No idea why!

ThunderCloud/GridPage.swift
simonmitchell: If we give the properties an initial value we get the error:
```
Immutable value 'self.dictionary' may only be initialized once
```
which is only fixable by changing the property to a `var` which I'd rather avoid for the moment :)

ThunderCloud/QuizGridItem.swift
simonmitchell: It is, because `badgeId` is a `let` 😝

simonmitchell:
@MattCheetham @ryanbourneuk changes made 👍

ThunderCloud/GridPage.swift
MattCheetham: Why not a var? If you make it a private var it will only be accessible inside the class.

ThunderCloud/QuizGridItem.swift
MattCheetham: Would suggest a private var again here with a default assignment of nil

ThunderCloud/GridPage.swift
simonmitchell: That's a good shout! Will do that 😆

ThunderCloud/QuizGridItem.swift
simonmitchell: I believe badgeId is used outside of the file here, so will try just making it internal for use in Storm 👍

simonmitchell: Actually I'm not sure I agree with this... I think badgeId shouldn't be made variable, even internally for the sake of removing a line of code. I really feel like if something does and should not change then it should be reflected in it's API, whether that's just internal to the framework or made public to the consumer 😊

simonmitchell: Never mind, look at me flip-flopping over here. it isn't used even outside this class so it can remain private var 😂

simonmitchell:
@MattCheetham good to go again 😊

PR #77, opened 2018-02-20
Sends `deviceId` to server when registering for push

MattCheetham:
Fixes a critical issue where ThunderCloud did not send `deviceId` as part of the payload when registering for push tokens. Unfortunately the server returns `202 Accepted` so this issue was not noticed sooner.


simonmitchell:
If this is needed, then obviously is approved :)

PR #79, opened 2018-03-08
Fixes HeaderListItemCell appearance

simonmitchell:
Looks like I accidentally left all the views in `HeaderListItemCell` disconnected from the `IBOutlets` on the cell file 🙃


simonmitchell:
Also updated to fix image height (So it's not constrained to the two text labels height) and to fix font sizing

PR #81, opened 2018-03-13
Makes sure `ListItem` correctly hides/un-hides text labels upon confi…

simonmitchell:
This fixes an issue where if a `ButtonListItem` was shown on certain size device on the same page as a `DescriptionListItem` the subtitle label would be hidden due to re-use issue in showing/hiding the labels. Showing and hiding of labels is now handled entirely by `ListItem`.

### DIFF
--- a/ThunderCloud/ListItem.swift
+++ b/ThunderCloud/ListItem.swift
@@ -58,10 +58,6 @@ open class ListItem: StormObject, Row {
 		if let linkDicationary = dictionary["link"] as? [AnyHashable : Any] {
 			link = StormLink(dictionary: linkDicationary)
 		}
-        
-        if title == "Timeframe to Appeal", let _subtitle = subtitle {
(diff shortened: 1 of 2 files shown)


ThunderCloud/ListItem.swift
```diff
@@ -87,6 +87,10 @@ open class ListItem: StormObject, Row {
 		
 		
 		if let tableCell = cell as? TableViewCell {
+            
+            tableCell.cellTextLabel?.isHidden = title == nil || title!.isEmpty
```
simonmitchell: Aware the indentation is off here... I think leave it for now and at some point we should go through the whole project and re-indent each file using Xcode so we don't see any more issues when moving from tabs -> Spaces

ryanbourneuk: Title seems to be force-unwrapped here?

PR #82, opened 2018-03-13
Fix/retain cycles

simonmitchell:
Whilst fixing a bug in a storm app I realised we have a fair few retain cycles being caused by various pieces of code. This fixes them by making circular references weak.


MattCheetham:
Code changes look fine. I'm going to assume you've seen this fix work in the memory graph :)

simonmitchell:
@MattCheetham I couldn't see the cycles in the memory graph but I debugged by putting a breakpoint in the `deinit` method of `ListPage`

PR #83, opened 2018-03-13
Changes showing quizzes from scroller to use same logic as

simonmitchell:
This changes the way quizzes are shown from `CollectionListItem` to replicate the behaviour of pushing a storm link from any other content. This enforces consistent behaviour rather than (on iPad in a split view) collection list item pushing the quiz rather than any other content replacing the detail VC

### DIFF
--- a/ThunderCloud/QuizBadgeScrollerViewCell.swift
+++ b/ThunderCloud/QuizBadgeScrollerViewCell.swift
@@ -68,6 +68,7 @@ open class QuizBadgeScrollerViewCell: CollectionCell {
 			} else {
 				parentViewController?.present(shareViewController, animated: false, completion: nil)
 			}
+            
 		} else {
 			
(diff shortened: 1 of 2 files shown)


ThunderCloud/QuizBadgeScrollerViewCell.swift
```diff
@@ -68,6 +68,7 @@ open class QuizBadgeScrollerViewCell: CollectionCell {
 			} else {
 				parentViewController?.present(shareViewController, animated: false, completion: nil)
 			}
+
```
ryanbourneuk: Is this blank line needed?

simonmitchell: Aye, it should be here because this is a multi-line if statement with an else below it ☺️

PR #87, opened 2018-03-26
Makes tweaks to YouTubeController

simonmitchell:
Minor tweaks to `YouTubeController` to fix slight parsing issue and to make sure only video streams that are valid are returned. 

Also adds mime type checking of streams via `AVURLAsset.audiovisualMIMETypes`.

### DIFF
--- a/ThunderCloud/YouTubeController.swift
+++ b/ThunderCloud/YouTubeController.swift
@@ -10,10 +10,15 @@ import Foundation
 
 public typealias YouTubeLoadCompletion = (_ url: URL?, _ error: Error?) -> Void
 
+/// A structural representation of a YouTube stream for a video.
 fileprivate struct YoutubeStream {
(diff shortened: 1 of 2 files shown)


simonmitchell:
Updated to add support for 720p and 1080p videos if clients so wish to add them.

ThunderCloud/YouTubeController.swift
```diff
@@ -10,6 +10,57 @@ import Foundation
...
+        }
+        var queryItems = components.queryItems ?? []
+        queryItems.append(URLQueryItem(name: "signature", value: signature))
+        components.queryItems = queryItems
+        return components.url
+    }
+    
+    let mimeType: String
+    
+    let codecs: [String]?
```
MattCheetham: Significant lack of documentation in this struct

PR #89, opened 2018-03-26
Fixes app being force into Portrait on iPad

simonmitchell:
Checks for supported app orientations before forcing the app back to portrait on all devices, this was causing some of our iPad apps to jump back to Portrait when dismissing videos

### DIFF
--- a/ThunderCloud/MultiVideoPlayerViewController.swift
+++ b/ThunderCloud/MultiVideoPlayerViewController.swift
@@ -8,6 +8,23 @@
 
 import UIKit
 
+fileprivate extension UIInterfaceOrientation {
+    init?(stringValue: String) {
+        switch stringValue {
+        case "UIInterfaceOrientationPortrait":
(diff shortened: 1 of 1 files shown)


ThunderCloud/MultiVideoPlayerViewController.swift
```diff
@@ -140,9 +157,30 @@ open class MultiVideoPlayerViewController: UIViewController {
...
+		
+        guard !supportedOrientations.contains(UIApplication.shared.statusBarOrientation) else {
+            return
+        }
+        
+		rotateDeviceToPortrait()
 	}
+    
+    private func rotateDeviceToPortrait() {
+
```
MattCheetham: So... this method is empty?

simonmitchell: @MattCheetham Mah bad, don't know what's wrong with me! Sorted

PR #93, opened 2018-04-10
Log upgrade

MattCheetham:
Adds a significant amount of debug logging to the delta download and update process. This should ensure easier debugging of issues in the future.

Error descriptions have been added for the content controller error enum and debug logs have been added to explain the process that is taking place. 

All of this can be accessed via the Console app

### DIFF
--- a/ThunderCloud/ContentController.swift
+++ b/ThunderCloud/ContentController.swift
@@ -651,6 +651,7 @@ public class ContentController: NSObject {
         var manifestJSON: Any
         
         // Serialize manifest into JSON
+        os_log("Loading manifest.json into JSON object", log: self.contentControllerLog, type: .debug)
         do {
             manifestJSON = try JSONSerialization.jsonObject(with: manifestData, options: JSONSerialization.ReadingOptions.mutableContainers)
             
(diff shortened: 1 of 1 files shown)


ThunderCloud/ContentController.swift
```diff
@@ -605,6 +617,7 @@ public class ContentController: NSObject {
 			do {
 				
 				// Remove unzip files
+                os_log("Cleaning up `data.tar.gz` and `data.tar` files", log: self.contentControllerLog, type: .debug)
```
ryanbourneuk: Guessing that this is caused by indentation type conflicts :)

ThunderCloud/ContentController.swift
```diff
@@ -661,86 +675,95 @@ public class ContentController: NSObject {
...
+        os_log("manifest.json exists", log: self.contentControllerLog, type: .debug)
         
         // Verify pages
+        os_log("Verifying pages", log: self.contentControllerLog, type: .debug)
         guard let pages = manifest["pages"] as? [[String: Any]] else {
             
-            os_log("No pages in manifest", log: self.contentControllerLog, type: .error)
-            callProgressHandlers(with: .verifying, error: ContentControllerError.invalidManifest)
+            os_log("%@", log: self.contentControllerLog, type: .error, ContentControllerError.missingPages.localizedDescription)
+            callProgressHandlers(with: .verifying, error: ContentControllerError.missingPages)
```
simonmitchell: Should this be called `missingPages` or `manifestMissingPages`? `missingPages` could be a bit confusing seeing as it could also refer to pages missing from the actual bundle 😊

ThunderCloud/ContentController.swift
```diff
@@ -1378,14 +1415,61 @@ enum ContentControllerError: Error {
...
     case noFilesInBundle
     case fileCopyFailed
-    case noTempDirectory
     case badFileRead
     case badFileWrite
     case defaultError
 }
+
+extension ContentControllerError: LocalizedError {
+    public var errorDescription: String? {
```
ryanbourneuk: Like this a lot :)

ThunderCloud/ContentController.swift
```diff
@@ -661,86 +675,95 @@ public class ContentController: NSObject {
...
             }
+            os_log("%@ exists in the bundle", log: self.contentControllerLog, type: .debug, source)
         }
         
         //Verify Content
+        os_log("Verifying Content", log: self.contentControllerLog, type: .debug)
         guard let contents = manifest["content"] as? [[String: Any]] else {
             
-            os_log("No content in manifest", log: self.contentControllerLog, type: .error)
+            os_log("%@", log: self.contentControllerLog, type: .error)
```
simonmitchell: There is nothing here for the `%@` to use

ThunderCloud/ContentController.swift
MattCheetham: yup

ThunderCloud/ContentController.swift
MattCheetham: Good point. Will rename.

PR #97, opened 2018-05-08
Fix/badge scroller cell

simonmitchell:
Fixes crashes and issues with badge scroller cell

### DIFF
--- a/ThunderCloud/CollectionCell.swift
+++ b/ThunderCloud/CollectionCell.swift
@@ -73,20 +73,22 @@ open class CollectionCell: StormTableViewCell {
 	}
 	
 	private func sharedInit() {
+        
+        guard let pageControl = pageControl, let collectionView = collectionView else { return }
 		
-		pageControl?.currentPage = 0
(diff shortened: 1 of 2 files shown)


ThunderCloud/CollectionCell.swift
```diff
@@ -75,19 +74,19 @@ open class CollectionCell: StormTableViewCell {
 	
 	private func sharedInit() {
 		
-		pageControl.currentPage = 0
-		pageControl.pageIndicatorTintColor = .lightGray
-		pageControl.currentPageIndicatorTintColor = ThemeManager.shared.theme.mainColor
-		pageControl.isUserInteractionEnabled = false
+		pageControl?.currentPage = 0
```
ryanbourneuk: Could `pageControl` and `collectionView` be unwrapped above, so we don't unwrap them inline for every time it's used? Not sure if it'd bring any performance benefits at all, but it'd save having to do it everywhere :)

simonmitchell: Done 👍

PR #98, opened 2018-05-09
Project upgrade

MattCheetham:
Upgrades all the frameworks to their latest versions and updates this framework to use Swift 4.1, removing redundant code and fixing any compiler warnings at the same time.

Any redundant commented out code was just removed. If the features need re-implementing we can do so later.


ThunderCloud/Language.swift
```diff
@@ -58,7 +58,7 @@ extension Language: Row {
         guard let currentLanguage = StormLanguageController.shared.currentLanguage, let languageId = languageIdentifier else { return UITableViewCellAccessoryType.none
         }
         
-        if let overrideLanguageId = StormLanguageController.shared.overrideLanguage?.languageIdentifier, overrideLanguageId == languageId {
+        if let overrideLanguageId = StormLanguageController.shared.overrideLanguagePack?.locale.identifier, overrideLanguageId == languageId {
```
simonmitchell: Is a locale identifier going to match a `Language.languageIdentifier`?

ThunderCloud/StormTableRow.swift
```diff
@@ -34,25 +34,6 @@ class StormTableRow: TableRow {
 				return
 			}
 			
-			//TODO: Make sure this still works!
```
simonmitchell: This was for RTL languages I believe, do we need to make sure they are still supported in Storm?

PR #102, opened 2018-06-12
Making the StormLanguageController open, to allow for unit testing.

ryanbourneuk:
Having a class marked as public prevents classes from outside of the module inheriting from it and overriding functionality - which is an issue for mocking classes during unit testing. I've changed this, although in the future I'm looking to see if there's a better way to go about this.

### DIFF
--- a/ThunderCloud/StormLanguageController.swift
+++ b/ThunderCloud/StormLanguageController.swift
@@ -16,13 +16,14 @@ import UIKit
 /// `Locale` comes in the format of "en_GB" (Language, Region)
 /// `Locale` is able to ingest locales in the three letter format provided they are in the language_region format.
 /// This controller often re-organises the Storm file names to be in the language_region format before converting to `Locale`, once these are converted to `Locale` they can easily be compared with `Locale`s from the users device to find a match.
-public class StormLanguageController: NSObject {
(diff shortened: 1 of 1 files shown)


ThunderCloud/StormLanguageController.swift
```diff
@@ -16,13 +16,14 @@ import UIKit
 /// `Locale` comes in the format of "en_GB" (Language, Region)
 /// `Locale` is able to ingest locales in the three letter format provided they are in the language_region format.
 /// This controller often re-organises the Storm file names to be in the language_region format before converting to `Locale`, once these are converted to `Locale` they can easily be compared with `Locale`s from the users device to find a match.
-public class StormLanguageController: NSObject {
```
simonmitchell: Whilst we're here, can we see if it's possible to remove the `NSObject` superclass?

MattCheetham: This is on a feature branch way back so it's unlikely it could be removed. Bit pointless as it's so far behind and we should do that change further upstream IMO

simonmitchell: Ahh fair, didn't realise that haha!

PR #107, opened 2018-10-24
Updating ThunderBasics framework, to use new MDCHUDActivityView code

ryanbourneuk:


simonmitchell:
Looks like tabs have been replaced with spaces @ryanbourneuk ? Can't remember what we decided on when we discussed this recently, but if we could re-indent the files in questions to they're at least consistent with either spaces **or** tabs that'd be awesome 😄

ryanbourneuk:
@simonmitchell, should be updated now to use the tabs indentation. Seems that for other things we're using spaces, but as we've seen in other PRs there's still some bits that are using tabs! All fixed now, thanks! :)

PR #108, opened 2018-10-25
Fixes user trying to use slider triggering back gesture

simonmitchell:
This fixes the issue in Storm where using the slider would trigger the back gesture of the `UIViewController`

### DIFF
--- a/ThunderCloud/QuizQuestionContainerViewController.swift
+++ b/ThunderCloud/QuizQuestionContainerViewController.swift
@@ -136,6 +136,16 @@ class QuizQuestionContainerViewController: UIViewController {
 		hintLabel.isHidden = question?.hint == nil
 		hintLabel.text = question?.hint
 	}
+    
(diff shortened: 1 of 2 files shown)


ThunderCloud/QuizQuestionContainerViewController.swift
```diff
@@ -136,6 +136,16 @@ class QuizQuestionContainerViewController: UIViewController {
...
 	}
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        navigationController?.interactivePopGestureRecognizer?.delegate = nil;
+    }
+    
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        navigationController?.interactivePopGestureRecognizer?.delegate = self;
```
ryanbourneuk: It'd be awesome to have this documented, just so it's super clear in the future why this is here :)

PR #111, opened 2018-11-13
Feature/thunderbasics swift

simonmitchell:
This is changes to facilitate the swiftification of large parts of ThunderBasics, but also brings in some fixes which I came across to existing code like "shake to localise"


ThunderCloud/NavigationController.swift
```diff
@@ -11,6 +11,7 @@ import SafariServices
 import MessageUI
 import StoreKit
 import ThunderTable
+import AVKit
```
ryanbourneuk: Is this necessary?

ThunderCloud/MultiVideoPlayerViewController.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import UIKit
+import AVKit
```
ryanbourneuk: Is this required?

ThunderCloud/YouTubeController.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import AVKit
```
ryanbourneuk: Is this import needed?

ThunderCloudTests/LocalisedStringTests.swift
```diff
@@ -0,0 +1,278 @@
...
+    }
+    
+    func testLowercased() {
+        let localised = "Test".localised(with: "basic_parameter_lowercased", paramDictionary: ["NAME": "SIMON"])
+        XCTAssertEqual(localised, "Hello simon")
+    }
+    
+    func testDate() {
+        let localised = "Test".localised(with: "date", paramDictionary: ["DATE": Date(timeIntervalSince1970: 0)])
+        XCTAssertEqual(localised, "It is Thu 01 January 1970 at 01:00")
```
ryanbourneuk: Is it 01:00 because of timezones?

ThunderCloudTests/LocalisedStringTests.swift
```diff
@@ -0,0 +1,278 @@
...
+    func testUnderline() {
+        
+        let underlined = "Test".localised(with: "underline", paramDictionary: ["LINK": "www.google.co.uk"])
+        XCTAssertEqual(underlined, "This is the link: www.google.co.uk")
+        
+        let attributedUnderlined = NSAttributedString().localised(with: "underline", paramDictionary: ["LINK": "www.google.co.uk"])
+        XCTAssertEqual(attributedUnderlined.string, "This is the link: www.google.co.uk")
+        
+        guard let range = attributedUnderlined.string.range(of: "www.google.co.uk") else {
+            return
```
ryanbourneuk: Would be good to get an `XCTFail` for the guard lets in this test fail :)

ThunderCloud/ListPage.swift
```diff
@@ -103,46 +104,52 @@ open class ListPage: TableViewController, StormObjectProtocol, TSCCoreSpotlightI
...
-		
-		return nil
-	}
+	
+}
+
+// MARK: - Core spotlight indexing
+extension ListPage: CoreSpotlightIndexable {
+    
+    public var searchableAttributeSet: CSSearchableItemAttributeSet? {
```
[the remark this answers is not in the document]
simonmitchell: Don't need the docs here, they will show via the docs on the protocol definition 😊

ThunderCloud/NavigationController.swift
simonmitchell: Yep, used on line 312 to create an `AVPlayer` :)

ThunderCloud/YouTubeController.swift
simonmitchell: Yep! 😊 Reason they've had to be added in is they used to be in the bridging header for ThunderBasics when it was objective-C!

ThunderCloud/MultiVideoPlayerViewController.swift
simonmitchell: See comments on other files! :D

ThunderCloud/LocalisedString.swift
```diff
@@ -26,19 +27,321 @@ public extension String {
...
+        case "uppercase", "uppercased":
+            return uppercased(with: currentLocale)
+        case "lowercase", "lowercased":
+            return lowercased(with: currentLocale)
+        case "capitalise", "capitalised", "capitalize", "capitalized":
+            return capitalized(with: currentLocale)
+        case "propercase", "propercased":
+            
+            // Lowercase to get rid of random uppercase letters
+            var returnString = lowercased(with: currentLocale)
```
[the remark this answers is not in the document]
simonmitchell: Yes, so returnString isn't being uppercased itself, calling `uppercased` returns the new string which is assigned to `testString`. The first character of each sentence is then pulled out of `testString` and replaced in the original `returnString` with each upper case letter!

ThunderCloud/EditLocalisationTableViewCell.swift
```diff
@@ -52,7 +52,7 @@ class EditLocalisationTableViewCell: InputTextViewCell {
 		super.layoutSubviews()
 		
 		backgroundView?.frame = CGRect(x: 8, y: 8, width: contentView.bounds.width-16, height: contentView.bounds.height - 8)
-		guard let cellTextLabel = cellTextLabel else {
+		guard let cellTextLabel = cellTextLabel?.superview else {
```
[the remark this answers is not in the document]
simonmitchell: It is yeah, because the label itself is in a container view which is the one which should have a RHS separator as defined in the lines below! Will rename the `let` now :)

PR #114, opened 2018-12-03
Makes SpotlightListItem fully subclass able

simonmitchell:
Makes SpotlightListItem's method  `spotlightCell(cell: SpotlightListItemCell, didReceiveTapOnItem atIndex: Int)` overridable

### DIFF
--- a/ThunderCloud/SpotlightListItem.swift
+++ b/ThunderCloud/SpotlightListItem.swift
@@ -9,7 +9,7 @@
 import UIKit
 import ThunderTable
 
-open class SpotlightListItem: ListItem {
+open class SpotlightListItem: ListItem, SpotlightListItemCellDelegate {
 	
 	/// An array of `Spotlight`s to be displayed
 	public var spotlights: [Spotlight]?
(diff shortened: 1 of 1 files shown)


ThunderCloud/SpotlightListItem.swift
```diff
@@ -65,18 +65,15 @@ open class SpotlightListItem: ListItem {
...
+    
+    open func spotlightCell(cell: SpotlightListItemCell, didReceiveTapOnItem atIndex: Int) {
+        
+        guard let spotlights = spotlights, spotlights.count > atIndex else { return }
+        let spotlight = spotlights[atIndex]
+        guard let link = spotlight.link else { return }
+        
+        self.link = link
+        parentNavigationController?.push(link: link)
+        NotificationCenter.default.sendStatEventNotification(category: "Spotlight", action: spotlight.link?.url?.absoluteString ?? "Unkown link", label: nil, value: nil, object: self)
```
MattCheetham: Ergh. Definitely should be "Unknown" but as it's been like this for a while, correcting it might break stats.

simonmitchell: 😞 such a dilemma 😔

PR #118, opened 2018-12-07
Fix/quix fixes

simonmitchell:
Makes numerous fixes to quizzes on translucent and non-dark navigation styled apps as well as removing a strange editable field on the completion screen, and an unused storyboard VC

### DIFF
--- a/ThunderCloud/AchievementDisplayView.swift
+++ b/ThunderCloud/AchievementDisplayView.swift
@@ -56,6 +56,8 @@ open class AchievementDisplayView: UIView, AchievementDisplayable {
 		subtitleLabel.textAlignment = .center
 		subtitleLabel.backgroundColor = .clear
 		subtitleLabel.font = titleLabel.font
+        subtitleLabel.isUserInteractionEnabled = false
+        subtitleLabel.isEditable = false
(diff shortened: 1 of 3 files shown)


ThunderCloud/AchievementDisplayView.swift
```diff
@@ -56,6 +56,8 @@ open class AchievementDisplayView: UIView, AchievementDisplayable {
 		subtitleLabel.textAlignment = .center
 		subtitleLabel.backgroundColor = .clear
 		subtitleLabel.font = titleLabel.font
+        subtitleLabel.isUserInteractionEnabled = false
```
ryanbourneuk: Indentation seems a bit strange here. Is that GitHub, or the file itself? :)

simonmitchell: Possibly the file itself, the whole of ThunderCloud needs a re-indent running at some point since we moved from tabs to spaces!

simonmitchell: @ryanbourneuk there we go! :P

ryanbourneuk: Thank you :D

PR #119, opened 2018-12-10
Checks for appId before launching dev mode login

simonmitchell:
Dev mode cannot be shown before we have an app ID, or app.json because the download will fail, and if the user switches back to live, they will have no content to be displayed if `app.json` is missing

### DIFF
--- a/ThunderCloud/DeveloperModeController.swift
+++ b/ThunderCloud/DeveloperModeController.swift
@@ -84,7 +84,7 @@ public class DeveloperModeController: NSObject {
         
         backgroundObserver = NotificationCenter.default.addObserver(forName: UIApplication.willEnterForegroundNotification, object: nil, queue: OperationQueue.main, using: { [weak self] (notification) in
             
-            // Make sure we have an app ID before entering developer mode, now that we have apps who download their bundle rather
+            // Make sure we have an app I and app.json before entering developer mode, now that we have apps who download their bundle rather
             // than being bundled with it.
             if DeveloperModeController.devModeOn, ContentController.shared.fileExistsInBundle(file: "app.json"),  UserDefaults.standard.string(forKey: "TSCAppId") ?? API_APPID != nil {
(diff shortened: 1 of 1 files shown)


ThunderCloud/DeveloperModeController.swift
```diff
@@ -84,7 +84,9 @@ public class DeveloperModeController: NSObject {
         
         backgroundObserver = NotificationCenter.default.addObserver(forName: UIApplication.willEnterForegroundNotification, object: nil, queue: OperationQueue.main, using: { [weak self] (notification) in
             
-            if DeveloperModeController.devModeOn {
+            // Make sure we have an app I and app.json before entering developer mode, now that we have apps who download their bundle rather
```
ryanbourneuk: `an app ID` instead of `an app I` :)

ThunderCloud/DeveloperModeController.swift
```diff
@@ -84,7 +84,9 @@ public class DeveloperModeController: NSObject {
...
+            // than being bundled with it.
+            if DeveloperModeController.devModeOn, ContentController.shared.fileExistsInBundle(file: "app.json"),  UserDefaults.standard.string(forKey: "TSCAppId") ?? API_APPID != nil {
```
ryanbourneuk: Wondering if `ContentController.shared.fileExistsInBundle(file: "app.json"),  UserDefaults.standard.string(forKey: "TSCAppId") ?? API_APPID != nil` would be better sat in a guard above this, with the `checkForUpdates` section on lines 95-97 being deferred? That way it'd be more explicit that before logging into dev mode / switching to live we must have those items.

Also, are we able to check for updates without those items? If so, a guard (and no defer) could be good? :)

simonmitchell: @ryanbourneuk that's a very good point... Have guarded now against all the code

PR #126, opened 2019-02-08
Fixes toolbar items when child of NavigationTabBarViewController

simonmitchell:
Fetches NavigationTabBarViewController's toolbarItems from the selected view controller!


ThunderCloud/NavigationTabBarViewController.swift
```diff
@@ -348,4 +348,11 @@ open class NavigationTabBarViewController: UIViewController, StormObjectProtocol
 			break
 		}
 	}
+    
+    open override var toolbarItems: [UIBarButtonItem]? {
```
ryanbourneuk: Some docs for this would be fab :)

simonmitchell: Have commented up why it's there! :D

PR #127, opened 2019-02-13
Fixing Video Controls layout

ryanbourneuk:
On newer devices, such as the iPhone X, the safe area insets were not taken into account when laying out views, resulting in the volume controls being unreachable (as they'd sit in the same region as the home bar).

Now, the safe area inset is taken into account, and on the newer devices the volume control is now usable!

### DIFF
--- a/ThunderCloud/MultiVideoPlayerViewController.swift
+++ b/ThunderCloud/MultiVideoPlayerViewController.swift
@@ -127,23 +127,32 @@ open class MultiVideoPlayerViewController: UIViewController {
 		
 		activityIndicator.startAnimating()
 	}
+    
+    private func getBottomSafeAreaInset() -> CGFloat {
+        if #available(iOS 11, *) {
+            return self.view.safeAreaInsets.bottom
+        } else {
+            return 0
+        }
+    }
 	
 	override open func viewWillLayoutSubviews() {
 		
 		super.viewWillLayoutSubviews()
 		
 		videoPlayerLayer?.frame = view.bounds
-		let orientation = UIApplication.shared.statusBarOrientation
+        
+        // Get the bottom safe area inset, so the controls can be shifted upwards on newer devices to remain accessible.
+        let bottomSafeAreaInset = getBottomSafeAreaInset()
+        
+        let orientation = UIApplication.shared.statusBarOrientation
 		
 		if orientation.isPortrait {
-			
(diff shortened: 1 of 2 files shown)


ThunderCloud/MultiVideoPlayerViewController.swift
```diff
@@ -127,23 +127,32 @@ open class MultiVideoPlayerViewController: UIViewController {
 		
 		activityIndicator.startAnimating()
 	}
+    
+    private func getBottomSafeAreaInset() -> CGFloat {
```
simonmitchell: Don't really need to worry about catering for sub-iOS 11 any more! Has the Deployment Target not been bumped yet? If not, then leave as-is :)

ThunderCloud/MultiVideoPlayerViewController.swift
```diff
@@ -127,23 +127,32 @@ open class MultiVideoPlayerViewController: UIViewController {
...
 	
 	override open func viewWillLayoutSubviews() {
 		
 		super.viewWillLayoutSubviews()
 		
 		videoPlayerLayer?.frame = view.bounds
-		let orientation = UIApplication.shared.statusBarOrientation
+        
+        // Get the bottom safe area inset, so the controls can be shifted upwards on newer devices to remain accessible.
+        let bottomSafeAreaInset = getBottomSafeAreaInset()
```
simonmitchell: re-indent dis file plz 😊

ThunderCloud/VideoPlayerControlsView.swift
```diff
@@ -45,10 +45,20 @@ open class VideoPlayerControlsView: UIView {
 	public required init?(coder aDecoder: NSCoder) {
 		super.init(coder: aDecoder)
 	}
+    
+    private func getBottomSafeAreaInset() -> CGFloat {
```
simonmitchell: Again, if our deployment target has been bumped, don't need this check :) Also again: re-indent :P

ThunderCloud/MultiVideoPlayerViewController.swift
ryanbourneuk: The target is still iOS 10 for this framework, I'll leave it as is - but then when it is bumped we'll go through and remove any  < iOS 11 handling code :)

ThunderCloud/MultiVideoPlayerViewController.swift
ryanbourneuk: Will do!

PR #131, opened 2019-03-12
Fixes incorrect table view reloading when quiz is completed

simonmitchell:
Fixes quiz reload not working / crashing caused by relying on `UINavigationController`'s visible view controller rather than the actual view controller containing the row!

### DIFF
--- a/ThunderCloud/ListItem.swift
+++ b/ThunderCloud/ListItem.swift
@@ -24,115 +24,115 @@ public protocol RowSelectable {
 /// ListItem is the base object for displaying table rows in storm.
 /// It complies to the `Row` protocol
 open class ListItem: StormObject, Row {
-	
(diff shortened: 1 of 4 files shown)


ThunderCloud/ListItem.swift
```diff
@@ -75,6 +78,8 @@ open class ListItem: StormObject, Row {
 	open func configure(cell: UITableViewCell, at indexPath: IndexPath, in tableViewController: TableViewController) {
 		
 		parentNavigationController = tableViewController.navigationController
+        parentViewController = tableViewController
```
ryanbourneuk: If you could reindent this file, that'd be fab :)

simonmitchell:
Sahweeeet, done! :D

PR #137, opened 2019-05-02
Feature/swift library upgrades

simonmitchell:
@ryanbourneuk a lot of this will be duplicated from the previous PR you approved, but happy for you to look through it all again with a fine tooth comb! The more we check this stuff, the safer it is!


ThunderCloud/LocalisationController.swift
```diff
@@ -445,7 +449,7 @@ public class LocalisationController: NSObject {
 			
 			// If any of the localised values (One for each language) has been edited
 			if localisation.localisationValues.first(where: { (localisationKeyValue) -> Bool in
-				return localisationKeyValue.localisedString == "".localised(with: localisationKey)
+				return localisationKeyValue.localisedString != "".localised(with: localisationKey)
```
ryanbourneuk: Is this intended?

ThunderCloud/TSCAppDelegate.swift
```diff
@@ -183,7 +185,8 @@ open class TSCAppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificatio
 	
 	public func setupSharedUserAgent() {
 		
-		TSCRequestController.setUserAgent(TSCStormConstants.userAgent())
+        //TODO: Add support back in!
```
ryanbourneuk: Can this be done yet?

ThunderCloud/StreamingContentController.swift
```diff
@@ -299,10 +305,11 @@ class StreamingContentFileOperation: CustomOperationBase {
...
         
-        fileRequestController.downloadFile(withPath: fileDownloadURLString, progress: { (progress: CGFloat, totalBytes: Int, bytesTransferred: Int) in
-            
-        }) { (fileLocation: URL?, downloadError: Error?) in
-            
+        guard let fileRequestController = fileRequestController else {
+            return
+        }
+        
+        fileRequestController.download("", progress: nil) { (response, fileLocation, error) in
```
ryanbourneuk: Is the first parameter of this meant to be empty?

ThunderCloud/StormNotificationHelper.swift
```diff
@@ -79,18 +80,20 @@ public class StormNotificationHelper {
...
 		
-		requestController.post("push/token", bodyParams: payload) { (response, error) in
-			
-			if error != nil {
-				return
-			}
-			
-			defaults.set(payload["token"], forKey: "TSCPushToken")
-		}
+//        requestController.post("push/token", bodyParams: payload) { (response, error) in
```
ryanbourneuk: Is this meant to be commented out?

ThunderCloud/LocalisationController.swift
simonmitchell: Yes, if `localisationKeyValue.localisedString != "".localised(with: localisationKey)` then we know the localisation has been edited and therefore colour the highlighted UI amber!

ThunderCloud/TSCAppDelegate.swift
simonmitchell: Not yet no, but I'll get on it now as it definitely needs doing!

ThunderCloud/StreamingContentController.swift
simonmitchell: Aye, the full URL to the file is passed into the init method of `StreamingContentFileOperation` so we just provide an empty string here! Will comment that up :)

PR #140, opened 2019-06-25
Added Settings.bundle to the framework's resources.

ryanbourneuk:
This should allow us to reference the Settings.bundle resource in apps, without needing to clone this entire repository as a submodule. This should speed up CI significantly! 🙌


simonmitchell:
Does this mean this will apply as the app's `Settings.bundle` automatically? Or will we still have to/ be able to reference it somehow? Only issue I can see here is for prep apps where we customise the `Settings.bundle` slightly ourselves to include user/device ID

ryanbourneuk:
@simonmitchell we'll still have to reference it from within the framework, it won't get automatically applied. It's just that now with it sat within this framework, we'll be able to reference it without also needing to have submodules :)

simonmitchell:
Ahhh so it's just accessible from within the compiled .framework?

PR #141, opened 2019-06-25
Fixes SpotlightListItem scroll crash

ryanbourneuk:
There was a crash in one of the apps, where there were no spotlights available when the automatic scroll timer fired and due to this, the app would crash.

This fixes this, by checking that there are spotlights and that the current page is within the correct bounds.


ryanbourneuk:
@simonmitchell, I've resolved your above comment in e23a029 :D

PR #147, opened 2019-07-09
Fixes ExternalLink navigation

ryanbourneuk:
When attempting to navigate to an ExternalLink, we must be referring to the main 'detail' UINavigationController (on an iPhone this is the top-most one, while on an iPad this is the detail navigation controller). Currently, it is referring to a different navigation controller (and so is presenting the SFSafariViewController incorrectly).

This PR resolves this, by adding an extension to fetch the top-most detail navigation controller, and uses that to present the web view.

### DIFF
--- a/ThunderCloud/NavigationController.swift
+++ b/ThunderCloud/NavigationController.swift
@@ -190,8 +190,12 @@ public extension UINavigationController {
 			UIApplication.shared.open(url, options: [:], completionHandler: nil)
 			
 		} else {
-            guard let rightMostNavigationController = UIApplication.shared.keyWindow?.rightMostNavigationController else {
-                return
+            var navigationController = self
+            
+            // Attempt to access the key window's right-most navigation controller.
+            // This resolves an issue on iPad where the SFSafariViewController is not presented correctly.
+            if let rightMostNavigationController = UIApplication.shared.keyWindow?.rightMostNavigationController {
+                navigationController = rightMostNavigationController
             }
 			
 			var url: URL?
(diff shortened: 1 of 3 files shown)


ThunderCloud/NavigationController.swift
```diff
@@ -202,13 +205,13 @@ public extension UINavigationController {
 			guard let _url = url else { return }
 			
 			let safariViewController = SFSafariViewController(url: _url)
-			safariViewController.delegate = self
+			safariViewController.delegate = rightMostNavigationController
```
simonmitchell: Unsure about setting the delegate to `rightMostNavigationController` here!

ThunderCloud/NavigationController.swift
```diff
@@ -190,6 +190,9 @@ public extension UINavigationController {
 			UIApplication.shared.open(url, options: [:], completionHandler: nil)
 			
 		} else {
+            guard let rightMostNavigationController = UIApplication.shared.keyWindow?.rightMostNavigationController else {
```
simonmitchell: Should we be falling back to `self` here?

ThunderCloud/NavigationController.swift
ryanbourneuk: So I'm of the thinking that if we're presenting on a different navigation controller, we should set the delegate to be that same navigation controller - it feels wrong that we'd be presenting on one and then handing delegate methods elsewhere. What do you think? :)

ThunderCloud/NavigationController.swift
ryanbourneuk: @simonmitchell resolved in 0677f8a8cab6c9fe6b2e2d17943394fb19e23f39 :D

ThunderCloud/NavigationController.swift
simonmitchell: Ahhh I didn't realise it didn't do anything with the delegate methods :D That's fine then!

PR #173, opened 2019-08-07
#172 - fixes embedded links not being presented.

### LINKED ISSUE #172: StormTableViewCell not displaying embedded links
## Expected Behaviour

When using a `ListItemView` object
```json
    "class": "ListItemView",
        "embeddedLinks": [
            {
                "class": "UriLink",
                "destination": "<link>",
                "title": { "class": "Text", "content": "<translation key>" }
            }
        ],
        "title": { "class": "Text", "content": "<translation key>" }
    }
```
the embedded links should be displayed. This would then match behaviour on Android.

## Current Behaviour

The `embeddedLinks` property is ignored when using `ListItemView`, and no links are presented.

## Possible Solution

To resolve this, the logic to handle embedded links should be added to the `StormTableViewCell` object. The `EmbeddedLinksListItemCell` class can then be removed, and any classes that previously inherited from it can inherit from `StormTableViewCell` instead.

1. Add the above JSON object into a `GroupView`'s children array, with valid translation keys and link.
2. Witness that the embedded link button is not shown.

## Context

Some call buttons within the GDPC First Aid apps are not being presented correctly.

ryanbourneuk:
## Description

This PR resolves #172 - where embedded links are not being presented when using the 
```json
"class": "ListItemView",
"embeddedLinks": [
    {
        "class": "UriLink",
        "destination": "<link>",
        "title": { "class": "Text", "content": "<translation key>" }
    }
],
"title": { "class": "Text", "content": "<translation key>" }
```
data structure.
## Related Issue

#172 

## Motivation and Context

Within the GDPC First Aid app, certain call buttons were not being correctly presented. This is in contrast to Android, where the buttons were correctly presented.

## How Has This Been Tested?

I have tested this, by making the change within the affected app's carthage submodule, and running the app with this change.

I have also added tests around `ListItem` & `ButtonListItem` to ensure functionality remains the same. This has involved a minor refactor of `StormLink` to accept injection of a `StormLanguageController` instance, however the existing tests continue to pass.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have added tests to cover my changes.


ryanbourneuk:
@simonmitchell `EmbeddedLinksListItem` has been removed! :D

ThunderCloud/StormLink.swift
```diff
@@ -12,239 +12,266 @@ import Foundation
...
+    
+    /// The body of the text to be shared
+    ///
+    /// Only valid for Email/SMS links
+    public var body: String?
+    
+    /// An array of recipients that the body should be shared to
+    ///
+    /// Only valid for SMS links
+    public var recipients: [String]?
```
ryanbourneuk: @simonmitchell many previous `let`s here have since been changed to use `var`. This hasn't caused any of the existing tests to fail, yet may be something to be aware of!

ThunderCloud/StormLink.swift
```diff
@@ -12,239 +12,266 @@ import Foundation
...
+            || linkClass == .emergency
+            || linkClass == .share
+            || linkClass == .timer
+            || linkClass == .external
+            || linkClass == .uri else {
+                return nil
+        }
+    }
+    
+    public func configure(with dictionary: [AnyHashable: Any], languageController: StormLanguageController) {
```
ryanbourneuk: @simonmitchell the changes in here (and this file) allow for the `languageController` instance to be injected so the title can be localised. This hasn't caused any of the existing tests to fail :)

PR #175, opened 2019-08-09
Fixes boolean logic around standard list item link classes!

simonmitchell:
## Description

This updates disclosure indicator logic on StandardListItem to match that from Obj-C version

## Motivation and Context

Solves disclosure indicators not being shown when they should be!

## How Has This Been Tested?

Has been tested by using this commit locally for BRC First Aid

## Types of changes

## Checklist:

### DIFF
--- a/ThunderCloud/StandardListItem.swift
+++ b/ThunderCloud/StandardListItem.swift
@@ -18,7 +18,7 @@ open class StandardListItem: ListItem {
 				return url.absoluteString.isEmpty ? UITableViewCell.AccessoryType.none : .disclosureIndicator
 			}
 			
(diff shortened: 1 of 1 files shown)


simonmitchell:
See old objC logic:
```
if (self.link.url) {
        
        if ([self.link.url.absoluteString isEqualToString:@""]) {
            return false;
        } else {
            return true;
        }
    } else if ([self.link.linkClass isEqualToString:@"SmsLink"] || [self.link.linkClass isEqualToString:@"EmergencyLink"] || [self.link.linkClass isEqualToString:@"ShareLink"] || [self.link.linkClass isEqualToString:@"TimerLink"]) {
        return true;
    } else {
        return false;
    }
```

PR #189, opened 2019-08-21
#188 - InlineButtonView timer button fix.

### LINKED ISSUE #188: Inactive timer text in InlineButtonView is incorrectly coloured.
## Expected Behaviour

The button title, when the timer is stopped, should be visible. This would align with the other buttons which do not present a `TimerLink` object.

## Current Behaviour

The button title colour is changed immediately upon the timer stopping to the same colour as the border. This, when in `solidMode` is the same as the background so the text is not visible. This only occurs when the `InlineButtonView` is displaying a `TimerLink` object.

## Possible Solution

Remove
```swift
if let borderColor = layer.borderColor {
  setTitleColor(UIColor(cgColor: borderColor), for: .normal)
}
```

1. Have a timer button within a cell.
2. Open the page that contains the cell
3. See that the button does not have text, click on the button to start the timer.
4. See that the timer countdown is visible
5. Click on the button to stop the timer
6. See that the text has vanished

## Your Environment

* ThunderCloud version: v1.4.0

ryanbourneuk:
## Description

Resolves #188, by removing the broken if-statement

## Motivation and Context

This change is required as without it, the timer text is invisible when the timer is stopped.

## How Has This Been Tested?

This has been run within the context of the affected project.

## Types of changes

## Checklist:

### DIFF
--- a/ThunderCloud/InlineButtonView.swift
+++ b/ThunderCloud/InlineButtonView.swift
@@ -63,11 +63,6 @@ open class InlineButtonView: TSCButton {
     
     /// Called when the button is representing a timer link and the user clicks to stop the timer or the timer elapses.
     open func stopTimer() {
-        
-        if let borderColor = layer.borderColor {
-            setTitleColor(UIColor(cgColor: borderColor), for: .normal)
-        }
-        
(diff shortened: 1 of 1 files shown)


simonmitchell:
No longer needed sorry @ryanbourneuk! (See issue)

PR #191, opened 2019-08-21
#190 - remove unneeded configure calls to remove duplicated buttons

### LINKED ISSUE #190: Duplicated Buttons within ButtonListItem
## Expected Behaviour

When showing a ButtonListItem, there should only be one button per link.

## Current Behaviour

Links are being duplicated & multiple buttons are appearing.

## Possible Solution

Remove duplicated configure calls, as these are automatically called upon `super.init(...)`

1. Have a ButtonListItem within a page, with a single button.
2. Navigate to that page in the app.
3. See that the button is duplicated.

## Your Environment

* ThunderCloud version: v1.4.0

ryanbourneuk:
## Description

Resolves #190, removing the duplicated buttons

## Related Issue

#190 

## Motivation and Context

Buttons should not be duplicated for each link.

## How Has This Been Tested?

It has been run in the context of the affected app, and unit tests have been updated to ensure that the right amount of links are generated.

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes.

### DIFF
--- a/ThunderCloud/ButtonListItem.swift
+++ b/ThunderCloud/ButtonListItem.swift
@@ -54,8 +54,6 @@ open class ButtonListItem: ListItem {
     /// - Parameter dictionary: A Storm object dictionary.
 	public required init(dictionary: [AnyHashable : Any]) {
 		super.init(dictionary: dictionary)
-		
-		configure(with: dictionary, languageController: StormLanguageController.shared)
 	}
     
     /// Given a Storm object dictionary, initialises a ButtonListItem.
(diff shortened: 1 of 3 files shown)


ThunderCloud/ButtonListItem.swift
```diff
@@ -54,8 +54,6 @@ open class ButtonListItem: ListItem {
     /// - Parameter dictionary: A Storm object dictionary.
 	public required init(dictionary: [AnyHashable : Any]) {
 		super.init(dictionary: dictionary)
-		
-		configure(with: dictionary, languageController: StormLanguageController.shared)
```
simonmitchell: Ahhhh, because it subclasses from `ListItem` now which does all this configure shizzle too!

PR #210, opened 2019-09-06
Feature/check view accessibility

### LINKED ISSUE #209: Update check views in content to use new image assets
## Expected Behaviour

Check views in content should use the new check view assets that quizzes now use

## Context

This is to make storm ADA compliant

simonmitchell:
## Description

Makes check views use the same image assets as quiz text selection items

## Related Issue

Resolves #209 

## Motivation and Context

Fixes checkable list item not using standard UI for selected state

## How Has This Been Tested?

Has been tested locally within ARC FA

IMG_0751

## Types of changes

## Checklist:

### DIFF
--- a/ThunderCloud/CheckView.swift
+++ b/ThunderCloud/CheckView.swift
@@ -12,17 +12,6 @@ import ThunderTable
 @IBDesignable
 class CheckView: UIControl {
 	
-	/// The corner radius of the view
-	@IBInspectable public var borderRadius: CGFloat {
-		get {
-			return layer.cornerRadius
-		}
-		set {
-			layer.cornerRadius = newValue
-			innerView?.cornerRadius = newValue
-		}
-	}
-	
 	/// The identifier for the check view.
 	///
 	/// This is used when saving the `CheckView`'s state to `UserDefaults` using `setOn:animated:saveState` so make sure if you do want to save the state you always provide the same identifier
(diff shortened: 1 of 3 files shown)


ThunderCloud/CheckView.swift
```diff
@@ -100,21 +54,8 @@ class CheckView: UIControl {
...
-		innerView?.cornerRadius = borderRadius - 3
-		innerView?.backgroundColor = .white
-		addSubview(innerView!)
-		
-		onTintColor = ThemeManager.shared.theme.mainColor
-		if tintColor == nil {
-			tintColor = UIColor(red: 0.90, green: 0.90, blue: 0.90, alpha: 1.0)
-		}
+        imageView = UIImageView(image: (#imageLiteral(resourceName: "check-off") as StormImageLiteral).image)
+		addSubview(imageView!)
```
ryanbourneuk: What'd (potentially) be amazing is an extension on UIView like `addOptionalSubview()` where you can pass in any optional UIView and it'll attempt to add it, otherwise it'll ignore it - so we don't have to do the force unwrapping. What are your thoughts on that? :)

simonmitchell: I’ll just add an `if let` in for now, but I like that as a suggested addition to ThunderBasics!

Issue #231, opened 2019-10-23
Make improvements to screen-reader behaviour


simonmitchell:
## Expected Behaviour

- ToggleableListItem should read as "Collapsable, Collapsed" e.t.c. based on it's state.
- App collection list should read out the "Installed" state of the app.
- Radio buttons should be read with role and selected state.
- Number of selected answers in quiz should be announced with context: '0 out of 1 possible answers selected'.
- Skip image area selection questions with announcement: 'This question cannot be completed with VoiceOver enabled, please navigate to next question'.
- Quiz slider position is announced as percentage rather than the value displayed above.

## Current Behaviour

None of the above happens

## Possible Solution

Make the fixes!

## Context

Need to make these changes to be ADA compliant

## Your Environment

* ThunderCloud version: 1.6.2
* iOS / macOS version: All
* Link to your project, if public: N/A

PR #233, opened 2019-11-13
#231 fix voiceover improvements

simonmitchell:
## Description
Makes improvements and fixes bugs with screen reader behaviour in Storm content

## Related Issue

#231 

## Motivation and Context

Fixes ADA compliance

## How Has This Been Tested?

This has been tested locally in multiple apps.

## Types of changes

## Checklist:

### DIFF
--- a/ThunderCloud/AppCollectionCell.swift
+++ b/ThunderCloud/AppCollectionCell.swift
@@ -26,6 +26,15 @@ extension AppCollectionItem: CollectionCellDisplayable {
         }
         return UIApplication.shared.canOpenURL(launchURL)
     }
+    
+    public var accessibilityLabel: String? {
+        let params = [
+            "APP_NAME": itemTitle ?? "Unknown".localised(with: "_APP_NAME_UNKNOWN")
+        ]
+        return enabled ?
+            "{APP_NAME}. Installed.".localised(with: "_APP_COLLECTION_ITEM_INSTALLED", paramDictionary: params) :
+            "{APP_NAME}. Not installed.".localised(with: "_APP_COLLECTION_ITEM_NOT_INSTALLED", paramDictionary: params)
+    }
 }
 
 /// A subclass of `CollectionCell` which displays the user a collection of apps.
--- a/ThunderCloud/EmbeddedLinksInputCheckItemCell.swift
+++ b/ThunderCloud/EmbeddedLinksInputCheckItemCell.swift
@@ -40,7 +40,7 @@ open class EmbeddedLinksInputCheckItemCell: StormTableViewCell {
     
     override open var accessibilityTraits: UIAccessibilityTraits {
         get {
-            return checkView.isSelected ? [.selected, .button] : [.button]
+            return checkView.isOn ? [.selected, .button] : [.button]
         }
         set { }
     }
@@ -61,7 +61,7 @@ open class EmbeddedLinksInputCheckItemCell: StormTableViewCell {
     
     override open var accessibilityHint: String? {
         get {
-            return checkView.isSelected ?
+            return checkView.isOn ?
                 "Selectable. Double tap to de-select".localised(with: "_CHECKITEM_VOICEOVER_HINT_SELECTED") :
                 "Selectable. Double tap to select".localised(with: "_CHECKITEM_VOICEOVER_HINT_UNSELECTED")
         }
--- a/ThunderCloud/EmbeddedLinksInputCheckItemCell.swift
+++ b/ThunderCloud/EmbeddedLinksInputCheckItemCell.swift
@@ -37,4 +37,34 @@ open class EmbeddedLinksInputCheckItemCell: StormTableViewCell {
 	required public init?(coder aDecoder: NSCoder) {
 		super.init(coder: aDecoder)
 	}
+    
+    override open var accessibilityTraits: UIAccessibilityTraits {
+        get {
+            return checkView.isSelected ? [.selected, .button] : [.button]
+        }
+        set { }
+    }
+    
+    override open var isAccessibilityElement: Bool {
+        get {
+            return true
+        }
+        set { }
+    }
+    
+    override open var accessibilityLabel: String? {
+        get {
+            return cellTextLabel?.text
+        }
+        set { }
+    }
+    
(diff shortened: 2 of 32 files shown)


ThunderCloud/AppCollectionCell.swift
```diff
@@ -26,6 +26,15 @@ extension AppCollectionItem: CollectionCellDisplayable {
         }
         return UIApplication.shared.canOpenURL(launchURL)
     }
+    
+    public var accessibilityLabel: String? {
+        let params = [
+            "APP_NAME": itemTitle ?? "Unknown".localised(with: "_APP_NAME_UNKNOWN")
+        ]
+        return enabled ?
+            "{APP_NAME}. Installed.".localised(with: "_APP_COLLECTION_ITEM_INSTALLED", paramDictionary: params) :
```
ryanbourneuk: Are there meant to be full stops after `{APP_NAME}`? Is this so when it's read out, there's a pause perhaps? :)

ThunderCloud/EmbeddedLinksInputCheckItemCell.swift
```diff
@@ -37,4 +37,34 @@ open class EmbeddedLinksInputCheckItemCell: StormTableViewCell {
...
+        get {
+            return cellTextLabel?.text
+        }
+        set { }
+    }
+    
+    override open var accessibilityHint: String? {
+        get {
+            return checkView.isOn ?
+                "Selectable. Double tap to de-select".localised(with: "_CHECKITEM_VOICEOVER_HINT_SELECTED") :
```
ryanbourneuk: Slightly insane how we don't get things like this for free from iOS! Is this because the entire cell needs the hint, instead of just a component within it? I remember at iOSDevUK they did cover this, might have been from a SwiftUI perspective mind...

ThunderCloud/List.swift
```diff
@@ -9,6 +9,28 @@
 import UIKit
 import ThunderTable
 
+class IndexableGroupView: StormObject, Section {
```
ryanbourneuk: Could be good to have some docs around this :)

ThunderCloud/QuizProgressListItemView.swift
```diff
@@ -9,11 +9,25 @@
 import UIKit
 import ThunderTable
 
+class IndexableQuizProgressListItemView: StormObjectProtocol {
```
ryanbourneuk: Docs around this too? :D

ThunderCloud/QuizQuestionContainerViewController.swift
```diff
@@ -148,8 +148,11 @@ open class QuizQuestionContainerViewController: AccessibilityRefreshingViewContr
         
         super.viewDidLoad()
 
-        navigationItem.rightBarButtonItem = UIBarButtonItem(image: (#imageLiteral(resourceName: "quiz-dismiss") as StormImageLiteral).image, style: .plain, target: self, action: #selector(handleQuitQuiz(_:)))
-        navigationItem.rightBarButtonItem?.accessibilityLabel = "Quit Quiz".localised(with: "_QUIZ_BUTTON_QUIT")
+        // If we're being presented and we're not the first view controller
```
ryanbourneuk: Comment says `&&`, logic says `||`. Is that intentional? :)

ThunderCloud/QuizSliderViewController.swift
```diff
@@ -36,6 +122,7 @@ class QuizSliderViewController: UIViewController, QuizQuestionViewController {
         if let image = question.sliderImage {
             
             imageView.accessibilityLabel = question.image?.accessibilityLabel
+            imageView.isAccessibilityElement = question.image?.accessibilityLabel != nil
```
ryanbourneuk: Within `ImageView`, could `isAccessibilityElement` be determined based on whether the accessibility label has been set? Might mean that all of these `!= nil` checks could be removed? :)

ThunderCloud/QuizSliderViewController.swift
```diff
@@ -54,13 +141,54 @@ class QuizSliderViewController: UIViewController, QuizQuestionViewController {
...
         slider.isUserInteractionEnabled = true
+        
+        sliderContainerView.isAccessibilityElement = false
+        sliderContainerView.accessibilityLabel = "Answer Slider".localised(with: "_QUIZ_VOICEOVER_SLIDER")
+        
+        guard let sliderContainerView = sliderContainerView else { return }
+        
+        sliderAccessibilityElement =             SliderAccessibilityElement(accessibilityContainer: sliderContainerView, dataSource: self)
+        sliderContainerView.accessibilityElements = [
+            sliderAccessibilityElement!
```
ryanbourneuk: Force unwrapping here - is this guaranteed to always be non-nil?

ThunderCloud/AppCollectionCell.swift
simonmitchell: @ryanbourneuk there is aye :D

ThunderCloud/EmbeddedLinksInputCheckItemCell.swift
simonmitchell: Yeah it's a PITA! as far as I could see there was no `selectable` accessibility trait provided by the OS :(

ThunderCloud/QuizQuestionContainerViewController.swift
simonmitchell: You are correct of course!

ThunderCloud/QuizSliderViewController.swift
simonmitchell: I would like to avoid this, as `ImageView` is provided by `ThunderTable` and therefore don't think we should force this logic upon users of that framework :) Especially as it's much more likely to be used by other devs!

simonmitchell: It's a great shout though... may in future subclass `ImageView` in ThunderCloud to do this, but pretty pressed for time atm :D

ryanbourneuk: That's a good shout actually, if it's provided by another framework then best leave it be for a potential future improvement :D

PR #237, opened 2019-11-28
Feature/badges

BenCShutt:
Updated the Badge model to include:
- campaign
- dateFrom
- dateUntil


ThunderCloud/DateFormatter+Extensions.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Copyright © 2019 threesidedcube. All rights reserved.
+//
+
+import Foundation
+
+extension DateFormatter {
+    
+    /// Standard `iso8601` `DateFormatter`
+    /// Note milliseconds including, for customization over `dateFormat`, see `iso8601(dateFormat:)`
+    static let iso8601: DateFormatter = {
```
timxyz: Can this call the `iso8601` function below, with `yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ` as format parameter?

simonmitchell: Like this suggestion @timxyz such Swift dev ;)

ThunderCloud/DateFormatter+Extensions.swift
```diff
@@ -0,0 +1,37 @@
...
+        formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ"
+        return formatter
+    }()
+    
+    /// `iso8601` `DateFormatter` with custom `dateFormat`.
+    /// This `DateFormatter` has:
+    /// - `.iso8601` `Calendar`
+    /// - "en_US_POSIX" `Locale`
+    /// - UTC `TimeZone`
+    static func iso8601(dateFormat: String) -> DateFormatter {
```
simonmitchell: There are functions for this in `ThunderBasics` @BenCShutt 😊 They don't set the `calendar` or `locale` on the date formatter however, so we may be best staying with this for now until we can (Perhaps optionally?) add those to `ThunderBasics`

ThunderCloud/DateFormatter+Extensions.swift
BenCShutt: Yeah makes a lot more sense! Oops

ThunderCloud/Badge.swift
```diff
@@ -8,8 +8,66 @@
...
+    case icon
+    case campaign
+    case dateFrom
+    case dateUntil
+}
+
 /// `Badge` is a model representation of a storm badge object
-open class Badge: NSObject, StormObjectProtocol {
+open class Badge: NSObject, StormObjectProtocol
+{
```
simonmitchell: Just noticed this @BenCShutt! 😊 Can we keep these in-line with what we've currently got with the opening brace on the same line? 😀 Not sure if there's maybe a setting in Xcode for it?

PR #242, opened 2019-12-03
Makes sure `".jpeg"` files in the bundle don't break the file exists check

simonmitchell:
## Description

Added `jpeg` into Boolean logic contained in `fileExistsInBundle` function in content controller.

## Motivation and Context

This caused an issue where if a delta update's `manifest.json` had a file in with the `.jpeg` extension, due to this file extension being missed in the `fileExistsInBundle` function if the file was moved to `Bundle.xcassets` by the `AppThinner` tool, then the bundle would incorrectly be marked as invalid!

## How Has This Been Tested?

Has been tested by pointing at this branch locally in the Blood project.

## Types of changes

## Checklist:

### DIFF
--- a/ThunderCloud/ContentController.swift
+++ b/ThunderCloud/ContentController.swift
@@ -12,6 +12,14 @@ import ThunderBasics
 import UIKit
 import os
 
+extension String {
+    /// Returns a Boolean value indicating whether the string contains any of the given elements.
+    /// - Parameter containedStrings: The substrings to check for
+    func containsOneOf<T: StringProtocol>(_ containedStrings: [T]) -> Bool {
+        return containedStrings.contains(where: { self.contains($0) })
+    }
+}
+
 let DOWNLOAD_REQUEST_TAG: Int = "TSCBundleRequestTag".hashValue
 
 // This needs to stay like this, it was a mistake, but without a migration piece just leave it be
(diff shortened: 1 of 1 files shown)


ThunderCloud/ContentController.swift
```diff
@@ -1256,10 +1264,12 @@ public extension ContentController {
         var thinnedAssetName = URL(fileURLWithPath: file).lastPathComponent
         let lastUnderScoreComponent = thinnedAssetName.components(separatedBy: "_").last
         
+        let extensions = [".jpg", ".jpeg", ".png"]
```
BenCShutt: Not sure if we want to `lowercase()` for safety, do you think that matters with paths?

ThunderCloud/ContentController.swift
```diff
@@ -1272,9 +1284,10 @@ public extension ContentController {
         if var imageSize = lastUnderScoreComponent {
             
             // Replace these for a later check
-            imageSize = imageSize.replacingOccurrences(of: ".jpg", with: "")
-            imageSize = imageSize.replacingOccurrences(of: ".png", with: "")
-            
+            extensions.forEach { (fileExtension) in
+                imageSize = imageSize.lowercased().replacingOccurrences(of: fileExtension, with: "")
```
BenCShutt: Extremely unlikely, but would break images with `.png` in their name. Could maybe ignore this one

ThunderCloud/ContentController.swift
```diff
@@ -12,6 +12,16 @@ import ThunderBasics
 import UIKit
 import os
 
+extension String {
+    /// Returns a Boolean value indicating whether the string contains any of the given elements.
+    /// - Parameter containedStrings: The substrings to check for
+    func containsOneOf(_ containedStrings: [String], caseSensitive: Bool = true) -> Bool {
```
BenCShutt: Nice :)

ThunderCloud/ContentController.swift
simonmitchell: It will I think... but I think we're safe to ignore for now!

PR #251, opened 2020-01-27
Quiz completion reformat

BenShutt:
Quiz completion text taken from localisations rather than quiz completion plugin.
Previously this text was on the `QuizCompletion` plugin model, now moved over to localisation.


simonmitchell (commented):
Have these properties been removed entirely from the Storm model? Or are we just ignoring them basically? :)

BenCShutt:
Fields: `cta` and `popup` have been removed from the `QuizCompletion` plugin (model). This should be across all apps, in preference of localisations. I am confirming now though

simonmitchell (approved):

PR #253, opened 2020-02-21
Fix/accessibility improvements

simonmitchell:
## Description

Makes general accessibility (Mostly VoiceOver) improvements to ThunderCloud

## Motivation and Context

Required for improved ADA compliance

## How Has This Been Tested?

Tested in a release build of one of our apps

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/ThunderCloud/QuizAreaSelectionViewController.swift
+++ b/ThunderCloud/QuizAreaSelectionViewController.swift
@@ -18,7 +18,13 @@ open class QuizAreaSelectionViewController: UIViewController, QuizQuestionViewCo
     
     fileprivate var circleLayer: CAShapeLayer?
     
-    open var circleColor: UIColor? = .white
+    fileprivate var circleShadowLayer: CAShapeLayer?
+    
+    fileprivate var circleInnerLayer: CAShapeLayer?
+    
+    open static var CircleInnerColor: UIColor = ThemeManager.shared.theme.mainColor
+    
+    open static var CircleOuterColor: UIColor = .white
     
     @IBOutlet weak var heightConstraint: NSLayoutConstraint!
     
@@ -33,10 +39,6 @@ open class QuizAreaSelectionViewController: UIViewController, QuizQuestionViewCo
         guard let question = question else {
             return
         }
-        
-        let imageAnalyser = ImageColorAnalyzer(image: question.selectionImage.image)
-        imageAnalyser.analyze()
-        circleColor = imageAnalyser.detailColor ?? .black
     }
     
     override open func viewDidAppear(_ animated: Bool) {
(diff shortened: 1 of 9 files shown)


ThunderCloud/QuizQuestionContainerViewController.swift
```diff
@@ -282,26 +282,49 @@ open class QuizQuestionContainerViewController: AccessibilityRefreshingViewContr
...
+            titleLabel.text = "\(questions.count) \("of".localised(with: "_QUIZ_OF")) \(quiz.currentIndex + 1)"
         } else {
-            progressLabel.text = "\(quiz.currentIndex + 1) \("of".localised(with: "_QUIZ_OF")) \(questions.count)"
+            titleLabel.text = "\(quiz.currentIndex + 1) \("of".localised(with: "_QUIZ_OF")) \(questions.count)"
         }
         
-        progressContainer.addSubview(progressLabel)
+        progressContainer.addSubview(titleLabel)
+        
+        let progress = Float(quiz.currentIndex) / Float(questions.count)
```
BenCShutt: Unlikely but could do with a `!= 0` check somewhere

ThunderCloud/QuizQuestionContainerViewController.swift
```diff
@@ -422,7 +445,36 @@ extension QuizQuestionContainerViewController: UIGestureRecognizerDelegate {
 extension QuizQuestionContainerViewController: QuizQuestionViewControllerDelegate {
     
     func quizQuestionViewController(_ questionViewController: QuizQuestionViewController, didChangeAnswerFor question: QuizQuestion) {
+        
         redrawSelectedLabel()
         redrawContinueButton()
+        
+        var selected: String?
```
BenCShutt: Why `String`?

ThunderCloud/QuizAreaSelectionViewController.swift
```diff
@@ -68,42 +66,55 @@ open class QuizAreaSelectionViewController: UIViewController, QuizQuestionViewCo
         let relativeLocation = CGPoint(x: location.x / imageView.bounds.width, y: location.y / imageView.bounds.height)
         question?.answer = relativeLocation
         
-        if let circle = circleLayer {
-            circle.removeFromSuperlayer()
-        }
+        [circleLayer, circleInnerLayer, circleShadowLayer].forEach({ $0?.removeFromSuperlayer() })
```
BenCShutt: Do we need to be removing and adding again?

ThunderCloud/QuizAreaSelectionViewController.swift
```diff
@@ -18,7 +18,13 @@ open class QuizAreaSelectionViewController: UIViewController, QuizQuestionViewCo
     
     fileprivate var circleLayer: CAShapeLayer?
     
-    open var circleColor: UIColor? = .white
+    fileprivate var circleShadowLayer: CAShapeLayer?
```
BenCShutt: Depending a bit on your answer to the remove-add question below. I think this is great opportunity to use a `lazy var` and configure it there. We can set the `path` later. 
I do this a little bit in `SeparatorTableViewCell` in Blood. 
If we are on a clock then tbh it's something we can look at another time.

ThunderCloud/QuizAreaSelectionViewController.swift
simonmitchell: Not 100% sure, just went with what was there previously... probably no need to, but also no need to *not* if you get what I mean? There's also a possibility this was added in previously because there was an issue if you didn't remove it... We'll never know (As there wasn't a comment there) :P

ThunderCloud/QuizQuestionContainerViewController.swift
simonmitchell: No reason really, will change it over to int if we wish :)

PR #255, opened 2020-02-21
Make `backgroundColor` of container view of `chevronImageView` `.clea…

BenCShutt:
- Make `backgroundColor` of container view of `chevronImageView` `.clear` in `ToggeableListItemCell`


simonmitchell:
I had to do this for my dark mode work too :)

PR #263, opened 2020-03-09
Fix/accordion tabbar navigation height

simonmitchell:
## Description

Fixes the height of the accordion navigation bar to match the RHS view controller in the split view

## Motivation and Context

It looks nicer!

## How Has This Been Tested?

Tested locally in Nordstrom

Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2020-03-09 at 15 14 38

## Types of changes

## Checklist:

### DIFF
--- a/ThunderCloud/AccordionTabBarItemTableViewCell.swift
+++ b/ThunderCloud/AccordionTabBarItemTableViewCell.swift
@@ -18,8 +18,10 @@ class AccordionTabBarItemTableViewCell: TableViewCell {
 	@IBOutlet weak var customTitleHeightConstraint: NSLayoutConstraint!
 	
 	@IBOutlet weak var viewControllerView: UIView!
-	
-	override func prepareForReuse() {
+	    
+    @IBOutlet weak var headerHeightConstraint: NSLayoutConstraint!
+    
+    override func prepareForReuse() {
 		customTitleView.subviews.forEach({$0.removeFromSuperview()})
 		viewControllerView.subviews.forEach({$0.removeFromSuperview()})
 	}
--- a/ThunderCloud/AccordionTabBarViewController.swift
+++ b/ThunderCloud/AccordionTabBarViewController.swift
@@ -112,6 +112,7 @@ public class AccordionTabBarItem: Row {
             view.removeFromSuperview()
         }
         
+        accordionCell.headerHeightConstraint.constant = navigationBarHeight
         accordionCell.customTitleHeightConstraint.constant = searchBar != nil ? navigationBarHeight + 12 : (titleView != nil ? navigationBarHeight : 0)
         
         viewController.view.frame = accordionCell.viewControllerView.bounds
--- a/ThunderCloud/AccordionTabBarViewController.swift
+++ b/ThunderCloud/AccordionTabBarViewController.swift
@@ -268,7 +268,7 @@ open class AccordionTabBarViewController: TableViewController, StormObjectProtoc
         tableView.contentInset = .zero
         
         tabBarItems = viewControllers.map({ (viewController) -> AccordionTabBarItem in
-            return AccordionTabBarItem(viewController: viewController, navigationBarHeight: navigationController?.navigationBar.frame.height ?? 44)
+            return AccordionTabBarItem(viewController: viewController, navigationBarHeight: navigationBarHeight)
         })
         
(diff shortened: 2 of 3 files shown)


ThunderCloud/AccordionTabBarItemTableViewCell.swift
```diff
@@ -18,8 +18,10 @@ class AccordionTabBarItemTableViewCell: TableViewCell {
 	@IBOutlet weak var customTitleHeightConstraint: NSLayoutConstraint!
 	
 	@IBOutlet weak var viewControllerView: UIView!
-	
-	override func prepareForReuse() {
+	    
+    @IBOutlet weak var headerHeightConstraint: NSLayoutConstraint!
+    
+    override func prepareForReuse() {
```
simonmitchell: Will re-indent this file once approved :)

ThunderCloud/AccordionTabBarViewController.swift
```diff
@@ -55,24 +55,26 @@ public class AccordionTabBarItem: Row {
...
     
     public var expanded: Bool = false
     
     public var isFirstItem: Bool = false
     
     public func height(constrainedTo size: CGSize, in tableView: UITableView) -> CGFloat? {
-        let height = expanded ? remainingHeight + 44 : 44
-        return isFirstItem ? height + 20 : height
+        let height = expanded ? remainingHeight + navigationBarHeight : navigationBarHeight
+        return isFirstItem ? height + UIApplication.shared.statusBarFrame.height : height
```
BenCShutt: I know the `statusBarFrame` is fetched different when using iOS 13 `scenes` but guess this is OK if we don't use that?

ThunderCloud/AccordionTabBarViewController.swift
```diff
@@ -311,14 +312,26 @@ open class AccordionTabBarViewController: TableViewController, StormObjectProtoc
         navigationController.view.sendSubviewToBack(navigationController.navigationBar)
     }
     
+    private var navigationBarHeight: CGFloat {
```
BenCShutt: Can we not use safe areas? Constraining to `superview` vs `safeArea` where required?

ThunderCloud/AccordionTabBarViewController.swift
simonmitchell: Yeah don't think we'd have to worry too much about that for now :) A lot of Storm will need a re-think once we start using SwiftUI anyways :D

ThunderCloud/AccordionTabBarViewController.swift
```diff
@@ -101,7 +101,7 @@ public class AccordionTabBarItem: Row {
         accordionCell.cellTextLabel?.textColor = contrastingColor
         accordionCell.cellImageView?.tintColor = contrastingColor
         
-        accordionCell.topConstraint.constant = isFirstItem ? 28 : 8
+        accordionCell.topConstraint.constant = isFirstItem ? UIApplication.shared.statusBarFrame.height : 0
```
BenCShutt: Not sure why `28` was used here and `20` above, do we need `8` padding?

BenCShutt:
Changes look good to me :), but would be interested to hear your response about constraining to `safeArea` and `superview` instead of using explicit sizes

ThunderCloud/AccordionTabBarViewController.swift
simonmitchell: Unsure if safe areas will satisfy our requirements... This is used to provide the height of one particular view within a stack view within the cells that make up the left hand panel of the accordion view. The navigation bar is completely hidden, so wouldn't really want to rely on safe areas (And wouldn't be sure exactly where to access it from e.t.c)

ThunderCloud/AccordionTabBarViewController.swift
simonmitchell: Previously this was the spacing between the top of the cell and the image/text (Kinda like padding) now it's more like margin, and the 8 points of vertical padding isn't really needed any more as we rely on aligning centre vertical with the parent (And also > constants of 0)

ThunderCloud/AccordionTabBarViewController.swift
BenCShutt: Sure :) Let's go with this then! :P

PR #266, opened 2020-04-09
Feature/background content updates

simonmitchell:
## Description

Adds various APIs and internal logic to allow for background update APIs including:
- Handling content-available notifications
- Scheduled BGAppRefreshTask

## Motivation and Context

This solves content not being immediately available to users of Storm apps!

## How Has This Been Tested?

Has been tested locally by Robby and Myself within the ARC First Aid app. It may not be perfect yet, but this needs to go into PR soon to be merged!

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


ThunderCloud/BundleDiagnosticsTableViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+        
+        navigationController?.navigationBar.prefersLargeTitles = false
+        title = "Bundle Information"
+        
+        redraw()
+    }
+    
+    // MARK: - Redrawing
+    
+    private func timestampOfBundle(in directory: URL?) -> TimeInterval? {
```
BenCShutt, line 31: Not an issue but does it need to accept optional?

ThunderCloud/BundleDiagnosticsTableViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+            // Create the request
+            let uuidString = UUID().uuidString
+            let request = UNNotificationRequest(identifier: uuidString, content: content, trigger: trigger)
+            
+            // Schedule the request with the system.
+            let notificationCenter = UNUserNotificationCenter.current()
+            notificationCenter.add(request, withCompletionHandler: nil)
+            
+            DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
+                exit(-1)
```
BenCShutt, line 165: Might need to brush up on my C, and not sure this really matters too much :'). But do we want to exit with success as this was the intended behavior. I.e. with a `0`

ThunderCloud/BundleDiagnosticsTableViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+        
+        redraw()
+    }
+    
+    // MARK: - Redrawing
+    
+    private func timestampOfBundle(in directory: URL?) -> TimeInterval? {
+        guard let directory = directory else { return nil }
+        let manifestURL = directory.appendingPathComponent("manifest.json")
+        guard let jsonObject = try? JSONSerialization.jsonObject(with: manifestURL) else {
```
BenCShutt, line 34: Personal preference would be to  use `Codeable` and `JsonDecoder` and throw on our errors, then the caller can silence the error if they want? This still works good though!

ThunderCloud/BundleDiagnosticsTableViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+        buildDateRow.accessoryView = buildDateString != nil ? .copyAccessoryView : nil
+        
+        var bundleDateString: String?
+        if let bundleTimestamp = timestampOfBundle(in: ContentController.shared.bundleDirectory) {
+            bundleDateString = dateFormatter.string(from: Date(timeIntervalSince1970: bundleTimestamp))
+        }
+        let bundleTimestampRow = StormDiagnosticsRow(title: "Bundle", subtitle: bundleDateString ?? "Unknown")
+        if let _bundleDateString = bundleDateString {
+            bundleTimestampRow.selectionHandler = { (_, _, _, _) in
+                UIPasteboard.general.copy(string: _bundleDateString)
```
BenCShutt, line 65: Might be worth a real quick toast to say what it's done?

README.md
```diff
@@ -85,6 +85,10 @@ And of course sometimes you'll want to have native content for an app, which is
 `StormGenerator.register(viewControllerClass: NativeViewController.self, forNativePageName: “native_page”)`
 There are multiple ways to override the native behaviour of Thunder Cloud, more of which can be seen in the [Storm Demo Project](https://github.com/3sidedcube/iOS-Storm-Demo)
 
+### Supporting Background Content Downloads
+
+There are two mechanisms for background content downloads
```
BenCShutt, line 90: Short and sweet 😂

ThunderCloud/BundleDiagnosticsTableViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+//  Created by Simon Mitchell on 08/04/2020.
+//  Copyright © 2020 threesidedcube. All rights reserved.
+//
+
+import UIKit
+import ThunderTable
+
+class BundleDiagnosticTableViewController: TableViewController {
+    
+    let dateFormatter = DateFormatter()
```
BenCShutt, line 14: I was going to leave a comment on this because `viewDidLoad` is still pretty clean! 
But thinking about it, sometimes things do happen in `UIViewControllers` before `viewDidLoad` is called (might not be the case here!). So `lazy` instantiation will protect against this not being fully initialised until `viewDidLoad`.

ThunderCloud/JSONString.swift
```diff
@@ -0,0 +1,29 @@
...
+//  JSONString.swift
+//  ThunderCloud
+//
+//  Created by Simon Mitchell on 03/04/2020.
+//  Copyright © 2020 threesidedcube. All rights reserved.
+//
+
+import Foundation
+
+extension String {
```
BenCShutt, line 11: Useful! :D

ThunderCloud/BundleDiagnosticsTableViewController.swift
BenCShutt: Sorry, see this is done!

ThunderCloud/gunzip.swift
```diff
@@ -0,0 +1,73 @@
...
+        guard readBytes > 0 else {
+            if readBytes < 0 {
+                throw readBytes == -1 ? GunzipError.decompressionFailed : GunzipError.unknownError
+            }
+            break
+        }
+        
+        let writtenBytes = CFWriteStreamWrite(cfWriteStream, buffer, Int(readBytes))
+        
+        guard writtenBytes <= 0 else {
```
BenCShutt, line 59: Very very petty, but does read a little bit better if you write
```
guard writtenBytes > 0 else {
    throw GunzipError.decompressionFailed
}
```
and let the `while` loop do its thing

ThunderCloud/gunzip.swift
```diff
@@ -0,0 +1,73 @@
...
+    
+    guard let writeStream = OutputStream(toFileAtPath: destination.path, append: false) else {
+        throw GunzipError.failedToCreateWriteStream
+    }
+    
+    let cfWriteStream = writeStream as CFWriteStream
+    // Open the write stream
+    guard CFWriteStreamOpen(writeStream) else {
+        throw GunzipError.failedToOpenWriteStream
+    }
```
BenCShutt, line 26: Do we want to `defer { CFWriteStreamClose(cfWriteStream) }` here in case there is an error before the below `defer`?

ThunderCloud/gunzip.swift
```diff
@@ -0,0 +1,73 @@
+//
+//  gunzip.swift
+//  ThunderCloud
+//
+//  Created by Simon Mitchell on 07/04/2020.
+//  Copyright © 2020 threesidedcube. All rights reserved.
+//
+
+import Foundation
+import zlib
```
BenCShutt, line 10: Seems like a good find! Definitely an easier to read than the previous code. I haven't really looked at this libraries docs and the below seems reasonable as I read through.

Cartfile
```diff
@@ -1,4 +1,5 @@
-github "3sidedcube/ThunderTable" == 1.3.6
-github "3sidedcube/ThunderCollection" == 1.2.2
-github "3sidedcube/ThunderBasics" == 1.5.0
-github "3sidedcube/ThunderRequest" == 2.1.4
+github "3sidedcube/ThunderTable" == 1.4.1
+github "3sidedcube/ThunderCollection" == 1.2.4
+github "3sidedcube/ThunderBasics" == 1.5.2
+github "3sidedcube/ThunderRequest" == 2.2.1
+github "3sidedcube/Baymax" "feature/plist-viewer-improvements"
```
BenCShutt, line 5: If you're happy, would be great to get a Baymax release out when it's ready

ThunderCloud/TSCAppDelegate.swift
```diff
@@ -11,30 +11,51 @@ import UserNotifications
...
-		window?.makeKeyAndVisible()
-		
+				
+        setupRootWindow()
 		setupSharedUserAgent()
         
+        if let remoteNotification = launchOptions?[.remoteNotification] as? [String : Any], let aps = remoteNotification["aps"] as? [AnyHashable : Any] {
+            
+            baymax_log("App was launched by remote notification:\n\(String(remoteNotification) ?? "Unable to Parse")", subsystem: Logger.stormSubsystem, category: "PushNotifications", type: .info)
+            let launchedByContentPush = aps.keys.count == 1 && aps["content-available"] as? Int == 1
```
BenCShutt, line 45: Why is there necessarily only 1 key in `aps` if `content-available`. Feel like the APNS payload can still add stuff to `aps` if they want? Or is this an agreed Storm format

ThunderCloud/BundleDiagnosticsTableViewController.swift
simonmitchell: We okay to leave this as-is for now @BenCShutt ? :D

BenCShutt: Yeah absolutely, still works fine!

ThunderCloud/JSONString.swift
simonmitchell: Going to leave as-is for now for time constraints :) Not sure about having a `throw` on `init` either, seems a bit strange!

ThunderCloud/gunzip.swift
simonmitchell: Yes, I may not do it as a defer though, as I'm not sure in which order two `defer` statements will be called!

ThunderCloud/JSONString.swift
BenCShutt: Sure :)! Btw I think `throws` in an `init` is fine when multiple errors can occur and nullability isn't enough :P

ThunderCloud/TSCAppDelegate.swift
simonmitchell: Will change!

ThunderCloud/JSONString.swift
simonmitchell: Ahh yeah that's a fair point!

PR #273, opened 2020-07-10
Release/v2.1.0

simonmitchell:
Most of this has already been PRd for the blended learning work, only the readme and cartfile have changed outside of that!


ThunderCloud/DateFormatter+Extensions.swift
```diff
@@ -11,13 +11,35 @@ import Foundation
 extension DateFormatter {
     
     /// Standard `iso8601` `DateFormatter`
-    /// Note milliseconds including, for customization over `dateFormat`, see `iso8601(dateFormat:)`
-    static let iso8601: DateFormatter = {
+    /// Note milliseconds included, for customization over `dateFormat`, see `iso8601(dateFormat:timeZone:)`
+    static var iso8601: DateFormatter {
```
BenCShutt, line 15: Maybe we could remove this shorthand `.iso8601 ` for now (sorry I know this is the release branch). Since these are now in `ThunderBasics` and would be referenced there

simonmitchell: Is the code in `ThunderBasics` identical do you know? :D If so, I'm happy to 🚮

PR #279, opened 2020-09-03
Feature/scroll offset memory

simonmitchell:
## Description

Updates `ThunderTable` to version which supports scroll offset memory and implements the scroll offset protocol required for this functionality on appropriate table view cells

## Motivation and Context

Issues were found with re-use of scrollable table view cells when testing Blood

## How Has This Been Tested?

Tested in Blood

## Types of changes

## Checklist:


ThunderCloud/CollectionCell.swift
```diff
@@ -236,3 +240,14 @@ extension CollectionCell : UICollectionViewDataSource {
         return collectionCell
 	}
 }
+
+extension CollectionCell: ScrollOffsetManagable {
```
BenCShutt, line 244: When conforming a `class` to a `protocol` perhaps it's best to declare the conformance on the `class` declaration (in the main file) rather than an `extension`.
Since class support inheritance, a subclass might want to override the protocol conformance implementation.
If done in an extension we won't be able to override (maybe with `@objc` - but meh)

In the case of `struct`, `enum` as there's no inheritance this approach is defo cleaner to read and write! :)

simonmitchell: That's a very good shout! Will do so now :)

BenCShutt (approved):

PR #281, opened 2020-09-08
Feature/localisation debugging

simonmitchell:
## Description

Adds a debug tool to storm section of Baymax which allows debugging localised strings in-app! This pre-pends the human readable string with the key in square brackets: "[_SOME_LOCALISATION_KEY] Hello World"

## Motivation and Context

This allows clients and PMs (And us) to tell with more ease whether a certain string is hard-coded or localised properly

## How Has This Been Tested?

Has been tested locally within Blood project

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/ThunderCloud/LocalisationsTableViewController.swift
+++ b/ThunderCloud/LocalisationsTableViewController.swift
@@ -11,6 +11,16 @@ import ThunderTable
 
 class LocalisationsTableViewController: TableViewController {
     
+    override func viewDidLoad() {
+        
+        super.viewDidLoad()
+        
+        navigationController?.navigationBar.prefersLargeTitles = false
+        title = "Localisations"
+        
+        redraw()
+    }
+    
     // MARK: - Redrawing
 
     func redraw() {
--- a/ThunderCloud/LocalisationsTableViewController.swift
+++ b/ThunderCloud/LocalisationsTableViewController.swift
@@ -37,7 +37,6 @@ class LocalisationsTableViewController: TableViewController {
     }
     
     func showQuitAppAlert() {
-        
         let alertController = UIAlertController(
             title: "Debugging Localisations Changed",
             message: "Please force-quit the app and restart it for the change to fully take effect.",
--- a/ThunderCloud/LocalisationsTableViewController.swift
+++ b/ThunderCloud/LocalisationsTableViewController.swift
@@ -12,19 +12,15 @@ import ThunderTable
 class LocalisationsTableViewController: TableViewController {
     
     override func viewDidLoad() {
-        
         super.viewDidLoad()
-        
         navigationController?.navigationBar.prefersLargeTitles = false
         title = "Localisations"
-        
         redraw()
     }
     
     // MARK: - Redrawing
 
     func redraw() {
-        
         let debugRow = StormDiagnosticsSwitchRow(
             title: "Debug Localisations",
             subtitle: "Enabling this will pre-pend the localisation key to all strings that are localised in the UI to make it easy to spot hard-coded strings",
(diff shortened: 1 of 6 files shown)


ThunderCloud/LocalisationsTableViewController.swift
```diff
@@ -0,0 +1,53 @@
...
+//  Copyright © 2020 threesidedcube. All rights reserved.
+//
+
+import UIKit
+import ThunderTable
+
+class LocalisationsTableViewController: TableViewController {
+    
+    override func viewDidLoad() {
+
```
BenCShutt, line 15: We need to have a word about you adding a space from the function to the `super` call! 🤣

ThunderCloud/LocalisedString.swift
```diff
@@ -44,6 +44,9 @@ public extension String {
     func localised(with key: String, paramDictionary: [String: Any]?) -> String {
         var string = NSAttributedString(string: self).localised(with: key, paramDictionary: paramDictionary).string
         string.localisationKey = key
+        if LocalisationsTool.showDebugLocalisations && !string.hasPrefix("[\(key)]") {
```
BenCShutt, line 47: @simonmitchell This code will get hit A LOT and I'm a little worried about this constantly checking the `UserDefaults`. I _thiiiiink_ the `UserDefaults` is `Dictionary` in memory which updates a file when mutated. But if there are any IO (or DB) operations at all we need to cache this flag in memory ourselves.

BenCShutt, line 47: Also, feel like I'm missing something obvious, but when the prefix check?

ThunderCloud/LocalisationsTableViewController.swift
simonmitchell: I have a habit, and not sure where it's come from exactly... of always including a white-space line if a method is over a certain size :P See the whole of this file... yet in `LocalisationsTool.swift` I've not put a white-space line in `launchUI(in:)`! Is there anything in the swift style guidelines about it?

BenCShutt (changes requested):

ThunderCloud/LocalisedString.swift
simonmitchell: I'll add in-memory caching, that's a great shout!

simonmitchell: So basically I originally added this into all 3 base string methods... It turns out they all end up calling the method on localised string... but then I also added the call to `StormLanguageController` itself... If any of that logic changes I thought it best to keep it safe so we don't break this functionality (So we still add the prefix at the call-site even if underlying methods are changed to not add it) at a future date. Happy to remove though if we wish and rely entirely on `StormLanguageController`?

ThunderCloud/LocalisationsTableViewController.swift
BenCShutt: Haha nothing I'm aware of, but maybe! 😛

ThunderCloud/LocalisedString.swift
BenCShutt: IMO it's nice to do it in one place if that's possible, and others reference that. 
But I'm happy to go with what you think is best here!

ThunderCloud/LocalisationsTool.swift
```diff
@@ -0,0 +1,41 @@
...
+    var displayName: String {
+        return "Localisations"
+    }
+    
+    func launchUI(in navigationController: UINavigationController) {
+        let tableViewController = LocalisationsTableViewController(style: .grouped)
+        navigationController.show(tableViewController, sender: nil)
+    }
+    
+    private static var _showDebugLocalisations: Bool?
```
BenCShutt, line 24: Nice, that protects us 😄

BenCShutt (approved):

ThunderCloud/LocalisedString.swift
simonmitchell: Let's leave it for now, can always clean up at a later time if we reckon this is unlikely to change :)

PR #288, opened 2020-09-25
Update `carthage-build.sh` for XCode 12

BenCShutt:
## Description
Fix `carthage` build script for XCode 12

### DIFF
--- a/carthage-build.sh
+++ b/carthage-build.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 # carthage-build.sh
 # Usage example: ./carthage-build.sh --platform iOS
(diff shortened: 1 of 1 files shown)


BenCShutt:
Before we merge this we need to think about which script we want and how we are going to approach this going forward

simonmitchell:
This simplified script is fine, as long as we're targeting the version of Xcode that we're compiling binary releases for :)

simonmitchell (approved):

BenCShutt:
@simonmitchell We should all be on the GM release now so guess it makes sense, but happy to keep the old script if it's doing more for us!

PR #289, opened 2020-09-29
Fix/collection list item masking

simonmitchell:
## Description

Fixes images in collection list item not being masked to their circular container

## Motivation and Context

Fixes UI glitches in certain apps which use images with non-zero alpha background colour

## How Has This Been Tested?

Tested in Swim

## Types of changes

## Checklist:

### DIFF
--- a/ThunderCloud.xcodeproj/project.pbxproj
+++ b/ThunderCloud.xcodeproj/project.pbxproj
@@ -1849,6 +1849,7 @@
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = arm64;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
 					"$(PROJECT_DIR)/Carthage/Build/iOS",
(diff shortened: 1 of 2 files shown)


ThunderCloud.xcodeproj/project.pbxproj
```diff
@@ -1849,6 +1849,7 @@
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = arm64;
```
BenCShutt, line 1852: If low LOW, we should get this into some `xcconfig` or something so it can be easily shared and reused for different projects. This is how you ran it on the simulator?

BenCShutt (approved):

BenCShutt:
Was this missed in the migration? Because I thought we caught everywhere `cornerRadius` was used, maybe I missed it?

ThunderCloud.xcodeproj/project.pbxproj
simonmitchell: This is how I got the project to compile using `Carthage` 👀

simonmitchell:
I don't believe so no @BenCShutt! I expect that other apps use images with transparent backgrounds, and Swim is the first app which doesn't that we have updated to `ThunderCloud v2.0.0` :)

ThunderCloud.xcodeproj/project.pbxproj
BenCShutt: Hmmm, I was able to get it compile, hence the `v3.0.0` binary created

BenCShutt:
I saw this issue I think too in ARC Emergency @simonmitchell

PR #292, opened 2020-09-30
Adds back in a default shadow to `CollectionItemViewCell`

simonmitchell:
## Description

This adds back in a shadow to collection item view cell that seems to have been removed at the time that blended learning was merged into develop. I can't recall any legitimate reason for this removal, so this PR is re-instating the exact same shadow that was present before

## Motivation and Context

Solves shadows not appearing where they should on `CollectionItemViewCell`

## How Has This Been Tested?

Tested running this branch locally in ARC First Aid

Simulator Screen Shot - iPhone SE (2nd generation) - 2020-09-30 at 13 32 51

## Types of changes

## Checklist:

### DIFF
--- a/ThunderCloud/CircleProgressView.swift
+++ b/ThunderCloud/CircleProgressView.swift
@@ -7,6 +7,7 @@
 //
 
 import UIKit
+import ThunderBasics
 
 /// Circular `UIView` which strokes a `progressPath` on a `backgroundPath`.
 open class CircleProgressView: UIView {
@@ -35,29 +36,13 @@ open class CircleProgressView: UIView {
         }
     }
     
-    /// The shadow colour, defaults to .lightGray
-    var shadowColor: UIColor = .lightGray {
-        didSet {
-            updateShadow()
-        }
-    }
-    
-    /// The opacity of the shadow, defaults to 0.75
-    var shadowOpacity: Float = 0.75 {
-        didSet {
-            updateShadow()
-        }
-    }
-    
-    /// The offset of the shadow, defaults to `.zero`
-    var shadowOffset: CGSize = .zero {
-        didSet {
-            updateShadow()
-        }
-    }
-    
-    /// The radius of the shadow, defaults to `3`
-    var shadowRadius: CGFloat = 3 {
+    /// The shadow components to be applied if `hasShadow` = true
+    var shadowComponents: ShadowComponents? = .init(
+        radius: 3,
+        opacity: 0.75,
+        color: .lightGray,
+        offset: .zero
+    ) {
         didSet {
             updateShadow()
         }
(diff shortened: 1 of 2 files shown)


ThunderCloud/CircleProgressView.swift
```diff
@@ -28,13 +28,41 @@ open class CircleProgressView: UIView {
...
-    /// Apply default shadow configuration
+    /// Apply shadow configuration based on `shadowColor`, `shadowOpacity`, `shadowOffset`, `shadowRadius` properties
     public var hasShadow = true {
         didSet {
             updateShadow()
         }
     }
     
+    /// The shadow colour, defaults to .lightGray
+    var shadowColor: UIColor = .lightGray {
```
BenCShutt, line 39: Don't we have `shadowComponents` in `ThunderBasics` now ?

simonmitchell: Yep, but I couldn't think of any nice way of making that work with the concept of `hasShadow` I don't really want to remove the `hasShadow` property as it's public and therefore that would be a breaking change... Because of the way `shadow: ShadowComponents` works (It's an extension) you can't have a `didSet` and therefore you can't store the current `ShadowComponents` for re-applying if the user toggles `hasShadow`... Any ideas?

simonmitchell: I'm being stupid! I'm with you now :D

ThunderCloud/CircleProgressView.swift
```diff
@@ -28,13 +29,25 @@ open class CircleProgressView: UIView {
         }
     }
     
-    /// Apply default shadow configuration
+    /// Apply shadow configuration based on `shadowComponents` property
     public var hasShadow = true {
```
BenCShutt, line 33: Think we can remove this and just use nullability of `ShadowComponents` ? 😃

BenCShutt: Ahh I see your last comment, breaking change... Hmm would just have to handle accordingly in `updateShadow` as you have.

simonmitchell: See my comment on your original comment, the only problem is if we do that it's technically a breaking change... and I'm not a massive fan of doing a major release for something so small after we've literally just made it to `v3.0.1` 😂

BenCShutt: Though as it's `CircleProgressView` wonder if anyone is using `hasShadow`...

BenCShutt (approved):

ThunderCloud/CircleProgressView.swift
BenCShutt: But yeah, more than happy to leave 😄

simonmitchell: Probably nobody, but I want to be strict on ourselves so if we do move to semantic versioning in the future then we're well-practiced at not causing ourselves any headaches :D

PR #295, opened 2020-10-01
Fix/collection list item padding

simonmitchell:
## Description

Refactors `CollectionListItem` and associated files slightly to include insets on the collection view, to act as padding in the table view cell

## Motivation and Context

Previously `CollectionListItem` relied on section padding to make it look appropriate amongst content, now the table view cell has "padding" itself, so doesn't rely on any external factors

## How Has This Been Tested?

Has been tested locally in ARC First Aid

Screenshot 2020-10-01 at 12 30 50

## Types of changes

## Checklist:

### DIFF
--- a/ThunderCloud/CollectionCell.swift
+++ b/ThunderCloud/CollectionCell.swift
@@ -148,12 +148,7 @@ open class CollectionCell: StormTableViewCell, ScrollOffsetManagable {
 		super.init(style: style, reuseIdentifier: reuseIdentifier)
 		
 		collectionViewLayout.scrollDirection = .horizontal
-        collectionViewLayout.sectionInset = .init(
-            top: Constants.sectionInsets.top,
-            left: 0,
-            bottom: Constants.sectionInsets.bottom,
-            right: 0
-        )
+        collectionViewLayout.sectionInset = Constants.sectionInsets
 		
 		collectionView = AccessibleCollectionView(frame: .zero, collectionViewLayout: collectionViewLayout)
 		contentView.addSubview(collectionView)
--- a/ThunderCloud/CollectionCell.swift
+++ b/ThunderCloud/CollectionCell.swift
@@ -140,6 +140,12 @@ open class CollectionCell: StormTableViewCell, ScrollOffsetManagable {
 		super.init(style: style, reuseIdentifier: reuseIdentifier)
 		
 		collectionViewLayout.scrollDirection = .horizontal
+        collectionViewLayout.sectionInset = .init(
+            top: CollectionItemViewCell.Constants.cellPadding.top,
+            left: 0,
+            bottom: CollectionItemViewCell.Constants.cellPadding.bottom,
+            right: 0
+        )
 		
 		collectionView = AccessibleCollectionView(frame: .zero, collectionViewLayout: collectionViewLayout)
 		contentView.addSubview(collectionView)
--- a/ThunderCloud/CollectionCell.swift
+++ b/ThunderCloud/CollectionCell.swift
@@ -110,6 +110,17 @@ public extension CollectionCellDisplayable {
 /// A subclass of `StormTableViewCell` which displays the user a collection view
 open class CollectionCell: StormTableViewCell, ScrollOffsetManagable {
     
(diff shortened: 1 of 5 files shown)


ThunderCloud/CollectionCell.swift
```diff
@@ -219,8 +222,8 @@ extension CollectionCell : UICollectionViewDelegateFlowLayout {
         guard let items = items, items.indices.contains(indexPath.item) else { return .zero }
         
         let item = items[indexPath.item]
-        let size = CollectionItemViewCell.size(for: item)
-        return CGSize(width: size.width, height: maxHeight)
+        let size = CollectionItemViewCell.size(for: item, includingVerticalPadding: false)
```
BenCShutt, line 225: Previously all the cells were the same height, is this no longer the case now then? 😄

ThunderCloud/CollectionItemViewCell.swift
```diff
@@ -165,21 +165,24 @@ open class CollectionItemViewCell: UICollectionViewCell {
     /// Calculates the size of the collection list item for the given item
     ///
     /// - Parameter item: The item which will be rendered
+    /// - Parameter includingVerticalPadding: Whether to include the cell's padding in the calculation or not
     /// - Returns: The size the items content will occupy
-    public class func size(for item: CollectionCellDisplayable) -> CGSize {
+    public class func size(
+        for item: CollectionCellDisplayable,
+        includingVerticalPadding: Bool = true
```
BenCShutt, line 172: Should we include the vertical padding now if it's not on the cell but instead on the section inset of the collectionView?

ThunderCloud/CollectionCell.swift
simonmitchell: No, it should be fine for them all to vary :D AFAIK collection views will pin the cells to the top rather than centering them vertically so it should be fine to have variable heights... it means we can reduce computation slightly as we can remove the `maximumHeight` property

ThunderCloud/CollectionItemViewCell.swift
simonmitchell: So yeah we have two options here:

1. Remove the padding from this entirely and add the padding to both the section inset and to the height calculation in the `Row` implementation (CollectionListItem)
2. Add `includeVerticalPadding` and only set it to `false` for the collection layout delegate method that needs the ACTUAL height of the cells.

2. Is what I went with, but on reflection, perhaps `1.` is better?

ThunderCloud/CollectionCell.swift
BenCShutt: Yeah that's true I believe 😄

ThunderCloud/CollectionItemViewCell.swift
BenCShutt: So I guess this is what I was thinking in my head.
- The `sectionInset` makes sense but is a new thing. 
- The `cellPadding` is part of the cell itself, but this should really be added in the xib. Looking at ThunderCloud it's not adding any spacing in the xib.

I think this solution is fine, but we should prob rename `cellPadding` to `sectionInset`

IMG_1153

simonmitchell: That makes total sense, I'll update `cellPadding` to be `sectionInset`!

BenCShutt: @simonmitchell Added a photo for clarity, took me ages with this new `.heic` file? :P

ThunderCloud/CollectionListItem.swift
```diff
@@ -177,8 +177,11 @@ open class CollectionListItem: ListItem {
         let itemSizes = cellItems.map({ CollectionItemViewCell.size(for: $0) }).sorted { (size1, size2) -> Bool in
```
BenCShutt, line 177: We could make this more readable like
```
let firstItem = cellItems?
    .map { CollectionItemViewCell.size(for: $0) }
   .sorted {  $0.height > $1.height }
   .first
```
but this is absolutely fine

BenCShutt (approved):

PR #307, opened 2020-11-05
Fixes icon image not being shared

RodrigoLondon:
Fixes icon image not being shared

## Description

Badges not being shared properly.  bagdeIcon is a StormImage, which UIActivityViewController doesn't know anything about, it can only work with things like String, and UIImage.
Badges are only shareable via messages and other share functionality when the user shares on the quiz completion screen. If the user shares it the badge is not included, only the share text. 
## Related Issue

1. Complete a quiz
2. Click share
3. Share the quiz with the badge icon displaying via messages
4. Return to ‘Test’ tab
5. Click completed badge icon
6. Click to share in the same way as test step 2

## Motivation and Context

1. The change is required so the user can share the badge icon image. 
2. Solves the issue of icon image not being shared. 

## Types of changes

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [ ] All new and existing tests passed.

### DIFF
--- a/ThunderCloud/QuizBadgeCollectionCell.swift
+++ b/ThunderCloud/QuizBadgeCollectionCell.swift
@@ -129,7 +129,7 @@ open class QuizBadgeCollectionCell: CollectionCell {
         let defaultShareBadgeMessage = "Test Completed".localised(with: "_TEST_COMPLETED_SHARE")
         var items: [Any] = []
         
-        if let badgeIcon = quizBadge.badge.icon {
+        if let badgeIcon = quizBadge.badge.icon?.image {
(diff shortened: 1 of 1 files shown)


BenShutt:
@RodrigoLondon Update the PR to go into `develop`, the `release/v3.1.1` PR will go into `master`

BenShutt:
Ideally would be good to search for something like `.icon` or `UIActivityViewController` and check this isn't present elsewhere. But this will be good to merge without that.

BenShutt (approved):

simonmitchell:
Nice catch both!

PR #313, opened 2021-01-12
Fixes a bug which would cause the app content to be missing

simonmitchell:
## Description

If an an app is migrated from one Storm App ID to another
under a version bump then the app would be blank on first launch.

This is caused because if `setupRootWindow` is called before
`ContentController.shared.appLaunched()` is called then the
app.json is loaded from the previous app's delta bundle, which
is then subsequently deleted by the content controller when it
detects an app version change.

This should all be okay and not result in any issues, due to these
calls being synchronous, but it seems that there's some unforseen
asynchronous process which means that the files are deleted by
the content controller before the content is shown on-screen.

Previously this would have only resulted in the user being shown the
same content as the latest delta update for the app in question,
but due to some underlying issue it was resulting in a blank app until
next launch at which point the previous App ID's delta is fully removed

## How Has This Been Tested?

Tested locally in ARC First Aid by manually calling `appLaunched` before we call `super.appDidFinish...` in the app delegate

## Types of changes

## Checklist:

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,5 +1,5 @@
 github "3sidedcube/ThunderTable" == 1.6.1
 github "3sidedcube/ThunderCollection" == 1.3.1
-github "3sidedcube/ThunderBasics" == 2.0.1
+github "3sidedcube/ThunderBasics" == 2.1.0
 github "3sidedcube/ThunderRequest" == 2.3.1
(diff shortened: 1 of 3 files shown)


BenShutt (approved):

BenShutt:
Typically, I think, setting up the window and root viewControllers is done right at the end of launch

PR #318, opened 2021-05-12
Feature/xc frameworks

BenShutt:
Remove Xcode build setting excluding `arm64` as an architecture.
Not sure this should have been here in the first place; as I understood it, the Carthage script handled the workaround?
Because of this, I've marked it as a bug fix rather than a functionality update.

I've noticed for the mean time we need to use `--no-use-binaries` while building the `xcframework`s until we update our binaries to be `xcframework`. 

## Motivation and Context
Move all apps we can over to `xcframework` using Carthage

## How Has This Been Tested?
In referencing apps and frameworks

## Types of changes

## Checklist:


simonmitchell (approved):

BenShutt:
Will be done in v4.0.0

PR #331, opened 2022-10-06
Update OS

jamesnjones:
Update OS version

Needed in order to update other projects


ThunderCloud/NavigationController.swift
```diff
@@ -17,29 +17,21 @@ import AVKit
...
+
     var navigationBarTintColor: UIColor? { get }
-    
-    @available(iOS, deprecated: 13.0, obsoleted: 14.0, message: "Please use `UINavigationBarAppearance` variables instead")
+
     var navigationBarBackgroundColor: UIColor? { get }
-    
-    @available(iOS, deprecated: 13.0, obsoleted: 14.0, message: "Please use `UINavigationBarAppearance` variables instead")
+
     var navigationBarTitleTextAttributes: [NSAttributedString.Key : Any]? { get }
```
3scsamdavis: Can we get rid of all these obsoleted vars now? 🤔

jamesnjones: yes we can actually, good shout :)

m0wo:
Were you able to get the test suite passing? When I tried they failed at `LocalisedStringTests`. Not sure why, but the `NSAttributedString`s being returned from the `.localised` call don't have any attributes

3scsamdavis (approved):

ThunderCloud/NavigationController.swift
```diff
@@ -18,30 +18,6 @@ import AVKit
...
-    @available(iOS, deprecated: 13.0, obsoleted: 14.0, message: "Please use `UINavigationBarAppearance` variables instead")
-    var navigationBarTintColor: UIColor? { get }
-    
-    @available(iOS, deprecated: 13.0, obsoleted: 14.0, message: "Please use `UINavigationBarAppearance` variables instead")
-    var navigationBarBackgroundColor: UIColor? { get }
-    
-    @available(iOS, deprecated: 13.0, obsoleted: 14.0, message: "Please use `UINavigationBarAppearance` variables instead")
-    var navigationBarTitleTextAttributes: [NSAttributedString.Key : Any]? { get }
-    
     @available(iOS 13.0, *)
```
BenShutt: Don't think these annotations are needed now

BenShutt (approved):

BenShutt:
I'm happy with the changes that have been made here, though I do suspect there is much more we could do to this library. It's very old and legacy though so would need to be thought about more. Having it compiling for the iOS versions it supports is the main thing

PR #333, opened 2023-01-16
🟠 Normal: Remove dev mode transition [Small]

3scsamdavis:
## What?
Removes the super sexy and on-trend (and I'm pretty sure long since removed by Apple) _**page curl transition**_ that was used when switching in and out of dev mode.

## Why?
SP-1622
This took some tracking down and a half.
BRC First Aid's navigation (nav and tab bars) would become buggy after switching between dev and live bundles. Navigation animations broke, large title preference would be ignored, and navigation elements would disappear (such as back buttons and the BRC logo).
I originally assumed this was to do with the dev theme applied by ThunderCloud, but the issue occurred switching in or out of dev mode - which led me to this transition.

I'm assuming - especially since this is a feature not accessed by users - that removing this transition altogether is fine, but please let me know team if you can see issues with this when applied to our fleet of Storm apps.

## Screen recordings

Issue:
https://user-images.githubusercontent.com/84126596/212673607-195aacc7-de5d-48e2-b974-3e6c9e27d791.mp4

Resolved:
https://user-images.githubusercontent.com/84126596/212673741-4f7ab559-e3b0-43f8-b593-a655afe54578.mp4

### DIFF
--- a/ThunderCloud/DeveloperModeController.swift
+++ b/ThunderCloud/DeveloperModeController.swift
@@ -239,19 +239,9 @@ public class DeveloperModeController: NSObject {
     ///
     /// If your root view controller is not a `TSCAppViewController` overriding this will be necessary
     open var refreshHandler: (_ devMode: Bool) -> (Void) = { (devMode) -> (Void) in
-        
         let appView = StormObjectFactory.createAppViewController()
-        
-        var viewOptions: UIView.AnimationOptions = devMode ? .transitionCurlUp : .transitionCurlDown
-        
         let devModeController = DeveloperModeController.shared
-        
-        guard let currentView = devModeController.appWindow?.rootViewController?.view else { return }
-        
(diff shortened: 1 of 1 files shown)


BenShutt:
It's so mad trying to track these things down. So well done Sam. 😆 

I'm happy with the changes, I'm just conscious it's a change to all Storm apps (which I don't like touching 😬).
But I think you're right that maybe this is a default behavior we _should_ change.

If we wanted to support both (making choosing a default quick and easy) we could have an `animated: Bool` somewhere or something that defaults to `false`. Apps that want it the old way can make it `true` 🤔 .
Fine to leave I think because we can roll back the PR if we neeeeed to.

BenShutt (approved):

m0wo (approved):

BenShutt:
Reckon this needs declining (or merging) @3SCSamDavis?
I did something around the time in a [PR here](https://bitbucket.org/3sidedcube/brc-firstaid-ios-client/pull-requests/115).

PR #344, opened 2025-10-24
Allow overriding of external browser web link handling

3scsamdavis:
## What?

- Moved all of the link handlers into their own functions so that the logic is a bit easier to read and understand.
- Added a new `webLinkHandler` to `StormGenerator` which allows the app using ThunderCloud to override the handling of links that should open in an external browser.

## Why?

In the Blood app, the user can set an in-app preference for their external browser. Links in Storm pages should also respect this preference, so allowing this override enables this.


neil3sidedcube (approved):

juwoncube (approved):
🙌🏽

